PR #191, opened 2026-05-07
chore: adding new issue_templates for bug and feature

kdinev:
Closes #  

### Additional information (check all that apply):
 - [ ] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] All relevant tags have been applied to this PR


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,176 @@
...
+  - type: dropdown
+    id: hosting-model
+    attributes:
+      label: Hosting Model
+      description: Which Blazor hosting model are you using?
+      options:
+        - Blazor Server
+        - Blazor WebAssembly
+        - Blazor Hybrid (MAUI)
+        - Blazor United (Interactive SSR)
```
damyanpetev: That's not really a hosting model and I can't actually even pin an option some of in the newer projects.
There is an Auto rendering mode, that's actually might be more important to know. 
Would it make sense to Rreplace this with project type instead?

Would add another picker for Blazor's render modes regardless.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,176 @@
...
+      description: Which browser(s) are you experiencing the issue in?
+      multiple: true
+      options:
+        - Chrome
+        - Firefox
+        - Safari
+        - Edge
+        - Other
+    validations:
+      required: true
```
damyanpetev, line 116: Also while at it make the browser optional or with a N/A value since it's not applicable to pure build/compilation errors and not relevant for MAUI I think

.github/ISSUE_TEMPLATE/bug_report.yml
damyanpetev: Actually if the old projects are still on offer (seem to be in VS) and questions can be conditional/chained, we could ask for a project type (with Web App being the default) and have the render modes picker be tied to that (and MAUI? haven't tested those) and hidden for the standalone Server/WASM where the render mode is not used/implicit

damyanpetev (approved):

PR #235, opened 2026-07-01
feat: sync updated source for 7.2 and cleanup

damyanpetev:
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [x] Changelog

## How Has This Been Tested?

 - [x] Unit tests
 - [x] Manual testing
 - [x] Automated e2e tests

### Test Configuration:
 - **.NET version**:
 - **Hosting model**: 
 - **Browser(s)**:
 - **OS**:

## Checklist:
 - [ ] All relevant tags have been applied to this PR

Closes #


damyanpetev:
Current issues I've noticed:
- Chat templates have a timing issue, depending on which some/all/none may be picked up by the render #238
- Highlight stories more often than not need a "refresh" to actually load in correctly with highlighted matches. The component is rendered and initialized and the `searchText` is set/reflected:
   image
   Thinking it's possible this is either some storybook oddity or if reproducible elsewhere - could be a timing issue with the slotted content on the WC itself. Usually a refresh on the page resolves that., which is still odd.

dkamburov:
What about the wc themes? Seems like they are missing, is this expected?
image

MayaKirova:
@dkamburov 
Did you do the steps for Building and Running Locally from the ReadMe?

kdinev:
I would like to get the code formatting fixed before merging. I can take care of that.

dkamburov:
Are we keeping the static web assets base path to be `_content/IgniteUI.Blazor` ?

```
<StaticWebAssetBasePath>_content/IgniteUI.Blazor</StaticWebAssetBasePath>
```

damyanpetev:
> I would like to get the code formatting fixed before merging. I can take care of that.

I would like the entire code formatting fixed, but not just in this PR, so it belongs in a separate one. Preferably, setting up `dotnet format` as pre-commit hook and running on the repo (with editor config adjustments as needed)

damyanpetev:
> Are we keeping the static web assets base path to be `_content/IgniteUI.Blazor` ?
> 
> ```
> <StaticWebAssetBasePath>_content/IgniteUI.Blazor</StaticWebAssetBasePath>
> ```

Yeah, at least for now I've strayed from changing any public-facing API in this repo/package. As it stands, the current setup is reference everything as you would with `IgniteUI.Blazor.Trial` or `IgniteUI.Blazor` - same namespaces and component names too. And you just swap out the package if you want to access components from another, but no change to existing code required.

We might need to change that as we move towards having the packages co-exist/depend on each other though.

> Edit: Hopefully, with some of the script loading updates we can remove the need for them entirely and have the old app.js remain as just a dummy for backwards compat this package can carry.
> Noting that because this will likely be left as the only package carrying the CSS themes still and I don't think we can "redirect" those content stuff, so changing this path will eventually impact the full `IgniteUI.Blazor`

MayaKirova (approved):

dkamburov:
Looks good to me, two minor things I guess:
1. Now we can get rid off gulp dep, can we? and also the ingest script in package.json.
2.  How about adding note to changelog that we can now use both of the packages, but this can be done with higher versions than 26.1.51(or maybe we'll tag 26.2 now?) for IgniteUI.Blazor and 1.0.0 for IgniteUI.Blazor.Lite

dkamburov (approved):

kdinev (approved):

PR #241, opened 2026-07-09
test: Adding API bulk automation to Blazor.Lite

MayaKirova:
## Description
Moving the existing bulk automation (with a few tweaks so they don't use metadata from the model) to public repo.

## Motivation / Context
So we reuse what we already have for automation.

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [x] CI/CD
 - [x] Tests
 - [ ] Changelog

### Component(s) / Area(s) Affected:
All component now have logic that checks all props/events/methods/templates for errors during blazor serialization/deserialization.

## Checklist:


.github/workflows/ci.yml
```diff
@@ -54,10 +54,27 @@ jobs:
...
           --results-directory ./TestResults
 
+      - name: Build Playwright test project
+        run: >
+          dotnet build ./tests/Infragistics.Blazor.Testing/Playwright.Testing/Playwright.Testing.csproj
+          --configuration Release
+
+      - name: Install Playwright
+        shell: pwsh
+        run: ./tests/Infragistics.Blazor.Testing/Playwright.Testing/bin/Release/net9.0/playwright.ps1 install
```
kdinev: Is there a dotnet10 playwright?

tests/Infragistics.Blazor.Testing/Playwright.Testing/Playwright.Testing.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
```
kdinev: Let's get all this updated to dotnet10.

kdinev (dismissed):

tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/Descriptions/Descriptions.csproj
damyanpetev, line 1: I the idea was to reflect off the actual components as they are, without needing the descriptions metadata and I don't even think it's used/available in this repo to begin with? So this project and its files should be redundant and can be removed?

.github/workflows/ci.yml
```diff
@@ -54,10 +54,27 @@ jobs:
...
+
+      - name: Run Bulk Automation
+        run: >
+          dotnet test ./tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing.sln
+          --settings ./tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/.runsettings
+          --no-build --configuration Release --verbosity normal
+          --collect:"XPlat Code Coverage"
```
damyanpetev: Might possibly want to figure out a meaningful name for this and possibly run those as separate projects?
```suggestion
      - name: Run Bulk API Automation
        run: >
          dotnet test ./tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing.csproj
          --settings ./tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/.runsettings
          --no-build --configuration Release --verbosity normal
          --collect:"API Code Coverage"
```
As we do have a separate Unit test project & I really don't know if we need the tests split into a solution of their own yet, or if we pull those into the main one. TBD

tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/CustomTypes.cs
```diff
@@ -0,0 +1,59 @@
...
+             { "LoadLayoutAsync", ["[]"] },
+        };
+
+        public static readonly Dictionary<string, object> PredefinedPropertyValues = new Dictionary<string, object>
+        {
+             { "Step", 0.5 },
+             { "Locale", "en" },
+            { "VisibleMonths", 2.0 },
+            // prompt needs to be a single symbol
+            { "Prompt", "/" },
```
damyanpetev, line 49: These are still better off scoped against components to avoid collisions in general (like/in the component config json).
Also could check if it's acceptable to annotate Blazor component parameters with `System.ComponentModel.DataAnnotations` stuff, specifically for things like the prompt where [`[MaxLength(1)]`](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.maxlengthattribute?view=net-10.0) can be useful. Blazor uses those for validation, just not sure if it'll mess with components.

Either way, both just as notes and can be ignored for now.

tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/ReflectionUtils.cs
```diff
@@ -0,0 +1,327 @@
...
+        public static List<string> excludedEvents = new List<string>() { };
+
+        // gets the full list of valid properties for type
+        public static List<PropertyInfo> GetValidProps(Type type)
+        {
+            List<string> basePropNames = (typeof(BaseRendererControl)).GetProperties(BindingFlags.Public | BindingFlags.Instance).Select(x => x.Name).ToList();
+            var props = type.GetProperties(BindingFlags.Public | BindingFlags.Instance)
+            // exclude readonly, broken component specific props, dependant props, props from the base renderer, templates and events
+            // exlude Name as it is used as unique identifier, we don't want to dynamically set it
+            .Where(x => x.CanWrite && x.Name != "Name" && !excludedProps.Contains(x.Name) && !dependantProps.Contains(x.Name) && !basePropNames.Contains(x.Name) && !x.Name.EndsWith("Script") &&
```
damyanpetev, line 27: Out of curiosity - can this be done simply through reflecting `[Parameter]` attributes?
I mean I don't think the framework `ComponentBase` adds any, and our `BaseRendererControl` does have a couple of `[Parameter]` but those should work, right?

Don't know if we have much in the way of settable non-param props and getter-only stuff end up as methods, right? The goal of course if to remove go without references of the old infrastructure (`BaseRendererControl`) so we can refactor components one by one while having the tests validate the switch

tests/Infragistics.Blazor.Testing/Blazor.TestBed.WebApp/Blazor.TestBed.WebApp/Components/Pages/Home.razor
```diff
@@ -0,0 +1,321 @@
...
+        ReflectionUtils.dependantProps = _app.config.GetSection(type + ":DependantProps")?.Get<string[]>()?.ToList() ?? new List<string>();
+        ReflectionUtils.dependantMethods = _app.config.GetSection(type + ":DependantMethods")?.Get<string[]>()?.ToList() ?? new List<string>();
+        ReflectionUtils.excludedEvents = _app.config.GetSection(type + ":ExcludedEvents")?.Get<string[]>()?.ToList() ?? new List<string>();
+
+        // wait for state change
+        _app.StateHasChanged();
+        await Task.Delay(1);
+
+        // wait for component ready
+        await (_app.dc.Instance as BaseRendererControl).EnsureReady();
```
damyanpetev: Hm, should make this optional and/or make our own alternative. Interesting if there's any guidance for client ready with Blazor Playwright testing

tests/Infragistics.Blazor.Testing/Blazor.TestBed.WebApp/Blazor.TestBed.WebApp/componentsConfig.json
```diff
@@ -0,0 +1,406 @@
+{
+  "IgbGrid": {
+    "ExcludedProps": [
+      // FilteringExpressionsTree and AdvancedFilteringExpressionsTree throw errors silently on the client and break all other props, haha :(
+      "FilteringExpressionsTree",
+      "AdvancedFilteringExpressionsTree",
```
damyanpetev: Pretty sure some of these are grid-related, so not really for this repo :)

damyanpetev (commented):
At least the project use nullable, which is great, but there seems to be some warnings, that'd be lovely if resolved (should be the standard for all new code).

Overall, looking to clean up any and all "leftover" bits that are either not needed/used/for this source and keep the test reflection as flexible as possible so we can refactor components one by one

We'll still need to adjust some thing as we move them around (such as `WCEnumNameAttribute` vs the System.Text.Json equivalent we'll likely move to, `BaseRendererControl` references, etc).

tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/ReflectionUtils.cs
MayaKirova: The ones from the BaseRendererControl still need to be excluded from the batch tests. Like these ones:

```

    [Parameter]
    public string Height
    {
        get; set;
    }

    [Parameter]
    public string Width
    {
        get; set;
    }

    [Parameter]
    public string Class
    {
        get; set;
    }

    [Parameter(CaptureUnmatchedValues = true)]    
    public Dictionary<string, object> AdditionalAttributes { get; set; }
```

They do work, in the sense that they add styles/classes/attributes to the component or its wrapper, but since the batch testing just checks whether a property set on the server is properly serialized and set to the client component and these are not actual properties on the client component - they need to be excluded for that test in some way.

tests/Infragistics.Blazor.Testing/Blazor.TestBed.WebApp/Blazor.TestBed.WebApp/Components/Pages/Home.razor
MayaKirova: If I remember correctly we do have our own handling for `EnsureReady` on the BaseRendererControl and it was checking whether the client component instance is initialized. Since we cannot start testing before the client component is initialized and can accept the props/methods/events etc.

.github/workflows/ci.yml
MayaKirova: I used the Playwright.Testing.csproj but left the solution file until we decide what to do with the test and test bed projects (add them to the main solution or leave them in separate solution).

tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/CustomTypes.cs
```diff
@@ -36,7 +37,6 @@ public static readonly Dictionary<string, object> PredefinedTypes
         {
              { "RegisterIconAsync", ["customIcon", "filter-solid.svg", "collection"] },
              { "RegisterIconFromTextAsync", ["customIcon2", "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 512 512\"><!--!Font Awesome Free 6.7.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2024 Fonticons, Inc.--><path d=\"M3.9 54.9C10.5 40.9 24.5 32 40 32l432 0c15.5 0 29.5 8.9 36.1 22.9s4.6 30.5-5.2 42.5L320 320.9 320 448c0 12.1-6.8 23.2-17.7 28.6s-23.8 4.3-33.5-3l-64-48c-8.1-6-12.8-15.5-12.8-25.6l0-79.1L9 97.3C-.7 85.4-2.8 68.8 3.9 54.9z\"/></svg>", "collection"]},
-             {"AddRowAsync", [ new NwindDataItem(){ }, null]},
              { "FocusPaneAsync", ["content1"] },
```
damyanpetev: Minor, but missed this one being probably DM method?

tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/.runsettings
```diff
@@ -0,0 +1,24 @@
...
+			<Headless>true</Headless>
+		</LaunchOptions>
+	</Playwright>
+	<TestRunParameters>
+		<Parameter name="defaultTimeout" value="40000" />
+	</TestRunParameters>
+
+	<DataCollectionRunSettings>
+		<DataCollectors>
+			<DataCollector friendlyName="XPlat Code Coverage">
```
damyanpetev: And missed that one

```suggestion
			<DataCollector friendlyName="API Code Coverage">
```

tests/IgniteUI.Blazor.Lite.IntegrationTests/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+- If you don't have `pwsh`, you will have to [install Powershell](https://docs.microsoft.com/powershell/scripting/install/installing-powershell)
+
+## 1. Build Infragistics.Blazor
+
+This project depends on the Infragistics.Blazor project. That is the base project that contains our Blazor components, which we're testing here.
+
+There's a detailed guide for building it [here](https://infragistics.visualstudio.com/NetAdvantage/_wiki/wikis/NetAdvantage.wiki/1107/Running-and-Debugging-Translator?anchor=debugging-blazor-project-and-source).
+
+Good Luck! You'll need it.
```
damyanpetev: Indeed! But hopefully not needed as much in this repo :)
Technically all you need to do is the npm steps in the root README.md

tests/IgniteUI.Blazor.Lite.IntegrationTests/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## 2. Install Playwright browsers
+
+- Open a Powershell terminal in the `./browser/IgBlazorSamples.Test/bin/Debug/net9.0` folder.
+
+- Run `./playwright.ps1 install chromium` to instal only chromium for Playwright to run. For all browsers just omit the  `chromium` argument.
+
+You can run it from the `./browser/IgBlazorSamples.Test` folder as well, but you will need to prefix the path to it: `bin/Debug/net9.0/playwright.ps1 install`
```
damyanpetev:
```suggestion
- Open a Powershell terminal in the `./test/Playwright.Testing/bin/Debug/net10.0` folder.

- Run `./playwright.ps1 install chromium` to instal only chromium for Playwright to run. For all browsers just omit the  `chromium` argument.

You can run it from the `./test/Playwright.Testing` folder as well, but you will need to prefix the path to it: `bin/Debug/net10.0/playwright.ps1 install`
```
though we might want to make those just `/net<version>/` so the instruction can remain evergreen.

Also the paths might need further updates, but think that is the project you use, right?

tests/IgniteUI.Blazor.Lite.IntegrationTests/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+For running from **cmd** you will need to use `pwsh ./playwright.ps1 install`.
+
+For other specific browsers refer to `./playwright.ps1 install --help`
+
+## 3. Build Infragistics.Blazor.Testing.sln
+
+- Open the **Infragistics.Blazor.Testing.sln** solution
+
+- Build the solution. Should successfully build all projects.
```
damyanpetev: Would've said swap the solution for project, but I think this step can be dropped entirely (the build is implied in the run description next) and looking at the CI, since not all use VS, `dotnet test` does the trick too and builds just fine.

tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/.runsettings
damyanpetev: Although I think I kept some run settings on the root repo level somewhere, not sure if those are can be reused

tests/IgniteUI.Blazor.Lite.IntegrationTests/ComponentTest.cs
damyanpetev, line 1: would not mind the file name matching `ComponentTest` :)

tests/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing/Infragistics.Blazor.Testing.sln
damyanpetev, line 1: For structure:

At this point, if the runsettings can be shared with the main solution, I'm thinking we can probably drop this one and its folder entirely. I want to update the root to .slnx anyway and use solution filters (.slnf) as needed.
TBH it should be fine to include those in the main solution even now.

As for the other two projects left - I'm thinking Playwright.Testing -> IgniteUI.Blazor.Playwright or IgniteUI.Blazor.PlaywrightTest (with or without dot).

I'd also add IgniteUI to the TestBed (since it is specific) and move both directly under /test instead of one folder down.

damyanpetev (approved):

PR #266, opened 2026-07-23
test(integration): Enable back Calendar and Rating Change events

### LINKED ISSUE #262: Calendar, Select and Rating `Change` events fail in API bulk automation
See https://github.com/IgniteUI/igniteui-blazor/pull/251#issue-4900390562 , specifically the last commit/comment:

> [test: fix path-bound reflection check](https://github.com/IgniteUI/igniteui-blazor/pull/251/changes/d3e141f67051e315526487d1497ae574ff7bac31) - Another thing that flew under the radar in https://github.com/IgniteUI/igniteui-blazor/pull/241 was some two-way event detection that was based on project paths - way too fragile. Reworked with cleaner reflection, except the old code had ignored a few Change events that Claude tested, found failing and suppressed. Tagging @MayaKirova since those were not covered before, but see why and if we could

As the notes in the config sate, for Calendar/Select:
>  There were errors : Event with name: Change did not fire.
and for the Rating it seems to mess with the state _after_ the events checks, which should be looked at:
> There were errors : Property value mismatch after setting. Prop name:  Value. Expected:100.0, but got the following on the client: 5

Note, that the logic before this was excluding all `Change` events on this type (Input and Text Area should have the same and those don't have errors and are not ignored elsewhere I can see, so kinda baffling why just those 3 fail.

Yes, those events are part of the two-way infrastructure for the components `Value`, but they are not the explicit `ValueChanged`. They should be covered, though if technically prohibitive in the API bulk, at least as e2e scenarios.

MayaKirova:
Enable back Calendar and Rating Change events. Add more details on why Select Change cannot be mocked ATM.
Closes #262

### DIFF
--- a/tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/ReflectionUtils.cs
+++ b/tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/ReflectionUtils.cs
@@ -109,7 +109,7 @@ namespace IgniteUI.Blazor.Lite.TestBed.Components.Common
             return result;
         }
 
-        public static Type? GetOriginEventDetailType(Type eventArgsType)
+        public static Type? GetOriginEventDetailType(Type eventArgsType, Type componentType)
         {
             var detailProp = eventArgsType.GetProperty("Detail", BindingFlags.Public | BindingFlags.Instance);
             if (detailProp == null)
@@ -121,6 +121,12 @@ namespace IgniteUI.Blazor.Lite.TestBed.Components.Common
             if (detailType == typeof(string) || detailType.IsPrimitive || detailType == typeof(DateTime))
             {
                 return detailType;
+            }
+
+            // Calendar detail type is Object, but it actually casts to DateTime, so we need to handle that case specifically.
+            if (detailType == typeof(object) && componentType == typeof(IgbCalendar))
+            {
+                return typeof(DateTime);
             }
 
             return null;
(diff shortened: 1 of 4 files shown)


tests/IgniteUI.Blazor.Lite.TestBed/Components/Common/ReflectionUtils.cs
```diff
@@ -123,6 +123,12 @@ public static List<string> Get2WayBindingEvents(Type componentType)
                 return detailType;
             }
 
+            // Calendar detail type is Object, but it actually casts to DateTime, so we need to handle that case specifically.
+            if (detailType == typeof(object) && componentType == typeof(IgbCalendar))
+            {
+                return typeof(DateTime);
+            }
```
damyanpetev: Assuming the Calendar detail is `object` because the client event will emit `Date | Date[]`, right? While that worked in the past, ideally we'd address that. Can't make use of unions just yet in C# while supporting older .NET versions, but we could have derived types that narrow down the event detail to either option, though that'd complicate the test reflection too. Or have Calendar-specific event type with both options, much like the component itself or guide users towards the ValueChanged/ValuesChnaged events instead and shudder this one entirely.

Either way, we should probably make the event detail cleaner in place of having component-specific logic in the tests reflection.


```suggestion
            // TODO: Calendar detail type is Object, but it actually casts to DateTime, so we need to handle that case specifically.
            if (detailType == typeof(object) && componentType == typeof(IgbCalendar))
            {
                return typeof(DateTime);
            }
```

damyanpetev (approved):
While not a huge fan of the component-specific checks, I've logged the one for the calendar and the value change to fit the rating I suppose is reasonable enough. The rest LGTM

PR #286, opened 2026-08-03
docs: fill in public API XML docs, types and fixes

damyanpetev:
Fills in the missing XML docs across the component wrappers, sourced from the Ignite UI Web Components metadata rather than written from scratch, then reviewed and adapted so they read as .NET API documentation. Comments that were already present but unusable — imported markdown, JavaScript vocabulary, a few statements outright wrong for Blazor — are fixed in the same pass.

Comment lines only: no code, signatures, attributes or member order were touched.

## Where the text comes from

| Source | Used for |
|---|---|
| `custom-elements.json` (CEM) | Component class summaries, properties, events, methods |
| Shipped `*.d.ts` typings | Interfaces the CEM omits (chat options and render contexts, filtering options, highlight navigation) and `@internal` members it strips (`HasDateParts`, `HasTimeParts`) |
| The wrappers themselves | Event-args and payload types, which have no upstream description at all — derived from the events that carry them, the serialized payload keys, and the upstream emit sites |
| Hand-written | Blazor-only surface: focus/blur, two-way binding callbacks, resource modules, `*Script` parameters |

## Filled in

- **Components** — class summaries, `[Parameter]` properties, events and methods, including both overloads of every sync/async pair (upstream documents only the async one) and the read-only value accessors such as `GetCurrentValue` and `GetTotal`.
- **Event args and payload types** — undocumented across the board and absent from the manifest, so each was derived from the events that raise it and the payload it actually carries.
- **`*Script` parameters** — public API with no documentation until now. Event handlers point at the event they replace, template renderers describe what they render, and both show the `igRegisterScript` registration.
- **Base types** — each now states which components it backs instead of standing unexplained.
- **`AddIgniteUIBlazor`** and the client resource modules, including when registering explicitly is actually needed.

## Fixed

- **Markdown that never rendered** — backticks, links and bullet lists became `<c>`, `<see langword="…"/>`, `<see href="…">` and `<list type="bullet">` with resolved enum crefs.
- **Web component vocabulary** — `igc-card` reads as `<see cref="IgbCard"/>`, `` `show()` `` as `<see cref="Show"/>`, "returns a promise that resolves to" as "returns", and DOM-only advice with no Blazor equivalent is gone.
- **Statements that were wrong here** — `Chip.Remove` claimed to return the removed chip though the event carries a `bool`; `DateTimeInput.Change` described a "checked state"; the calendar range strings were described as labels when they are placeholders; "Cancelable." was dropped from `-ing` events, since Blazor's async round trip cannot cancel.
- **Structure** — return values moved into `<returns>` with consistent phrasing, and cross references point at the component a consumer uses rather than an internal base type.

## Verification

- `dotnet build` clean: **0 errors, 0 doc-syntax warnings** — no malformed XML (CS1570), no parameter mismatches (CS1572/1573), no unresolved or ambiguous crefs (CS1574/CS0419).
- Missing-doc warnings (CS1591) drop substantially. What remains is wrapper plumbing (constructors, `Type`, `FindByName`, serialization overrides, render flags) and the enums, all deliberately left alone pending future updates.

## Defects found while documenting — not fixed here

Documenting the API surface surfaced these; each needs a code change and its own decision:

1. **`IgbDropdown.Toggle(Object)` never toggles** — invokes `"show"` instead of `"toggle"`, so the sync overload can only open the dropdown. https://github.com/IgniteUI/igniteui-blazor/pull/288
2. **`AddIgniteUIBlazor(settings)` discards modules** — with no `modulesToLoad` params, it calls `WithModulesToLoad(null)` and clears whatever the settings carried. https://github.com/IgniteUI/igniteui-blazor/pull/312
3. **Wrapper defaults disagree with upstream** in a couple of dozen places — `Tooltip.ShowDelay` is `0` where upstream is `200`, `Textarea.Rows` `0` vs `3`, `SpinLoop` `false` vs `true`, `Tile.ColSpan`/`RowSpan` `0` vs `1`. Behavior is correct — a property that was never assigned is not serialized, so the component keeps its own default — but reading the property returns the wrong value. #292 
4. **`IgbDateRangePickerResourceStrings` has no properties** — upstream declares its own plus the inherited calendar strings, so the picker cannot be localized through that parameter. https://github.com/IgniteUI/igniteui-blazor/pull/295
5. **Dead surface** — `IgbCalendarDate` / `IgbCalendarDateEventArgs` have no upstream counterpart and nothing references them; `CalendarResourceStrings.SelectedDate` is mapped to nothing upstream and has no effect. https://github.com/IgniteUI/igniteui-blazor/pull/310


src/componentsBase/WebInputs/Tab.cs
```diff
@@ -5,6 +5,11 @@ namespace IgniteUI.Blazor.Controls
     public partial class IgbTab : BaseRendererControl, IDisposable
     {
         private EventCallback<bool>? _selectedChanged = null;
+
+        /// <summary>
+        /// Emitted when the Selected property changes.
+        /// Enables two-way binding through <c>@bind-Selected</c>.
+        /// </summary>
```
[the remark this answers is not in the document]
damyanpetev: Oh, cool, one more issue unlocked. To address separately though.

src/components/Blazor/Accordion.cs
```diff
@@ -126,13 +130,25 @@ public async Task ShowAllAsync()
...
 
         private string _openingRef = null;
         private string _openingScript = null;
+
+        /// <summary>
+        /// Name of a client-side function that handles the <see cref="Opening"/> event in the browser instead.
+        /// </summary>
+        /// <remarks>
+        /// Register the function on the client like
+        /// <c>igRegisterScript("MyHandler", function (args) { }, false)</c>.
```
dkamburov, line 150: this can add blazor code as well OpeningScript="MyHandler" as well, but I guess it go without it

dkamburov: I guess it can stay that way as well

dkamburov (approved):

src/components/Blazor/Accordion.cs
damyanpetev: Yeah, I assume that's obvious. The register call isn't so I went with the noisy description.

src/componentsBase/WebInputs/Tab.cs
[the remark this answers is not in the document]
damyanpetev: Covered in https://github.com/IgniteUI/igniteui-blazor/pull/315

PR #291, opened 2026-08-03
docs: Add calendar resource strings documentation.

### LINKED ISSUE #417: Document component-specific public API members
- [x] Document all public API members (properties, events, methods) #286, #291, #345, #422
- [x] Document components enum types #421

damyanpetev:
## Description
Addition to #286  for the calendar resource strings, since those need to be reviewed specially and/or updated.

 - [ ] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog

## Checklist:


src/components/Blazor/CalendarResourceStrings.cs
```diff
@@ -17,6 +21,10 @@ public IgbCalendarResourceStrings() : base()
         private string _selectMonth;
 
         partial void OnSelectMonthChanging(ref string newValue);
+
+        /// <summary>
+        /// Text for the accessible label of the header button that switches the calendar to the months view.
```
skrustev, line 26: Maybe more clear would be to say it is for the "accessibility attribute `aria-label`", since label might mean something that's visible?

Same for `SelectYear`.

src/components/Blazor/CalendarResourceStrings.cs
```diff
@@ -136,6 +175,11 @@ public string EndDate
         private string _previousMonth;
 
         partial void OnPreviousMonthChanging(ref string newValue);
+
+        /// <summary>
+        /// The label of the navigation button that moves the days view one month back.
```
skrustev, line 180: Also "aria-lable" related.
Same for `NextMonth`, `PreviousYear`, `NextYear`, `PreviousYears`, `NextYears`.

src/components/Blazor/CalendarResourceStrings.cs
```diff
@@ -51,6 +63,11 @@ public string SelectYear
         private string _selectDate;
 
         partial void OnSelectDateChanging(ref string newValue);
+
+        /// <summary>
+        /// Title shown in the calendar header until a date is selected, in single selection mode.
```
skrustev, line 68: So we decide on this to be the correct functionality, since currently it is always showing `Select Date`/`Select Range`? Probably should log it?

damyanpetev: I have not, which is why this is WIP and I've no idea what to do with those props  yet. Kinda why  I tucked this PR to the side since we need to clear what's the functionality behind these, if any

skrustev (dismissed):

src/components/Blazor/CalendarResourceStrings.cs
skrustev: After discussing it we've decided that we will not change the current behavior and we will just deprecate the SelectedDate property.

src/components/Blazor/CalendarResourceStrings.cs
```diff
@@ -2,13 +2,20 @@
...
+    /// </summary>
     public partial class IgbCalendarResourceStrings : BaseRendererElement
     {
         /// <inheritdoc />
         public override string Type { get { return "WebCalendarResourceStrings"; } }
 
         private string _selectMonth;
 
+        /// <summary>
+        /// The accessibility attribute <c>aria-label</c> of the header button that switches the calendar to the months view.
```
damyanpetev: @skrustev "[Accessible name](https://developer.mozilla.org/en-US/docs/Glossary/Accessible_name)" would the technically correct term, but to avoid relying on users being familiar w/ the ARIA glossary, went with using "label" instead. However, that was still in general term, without disclosing actual implementation (yes it's `aria-label` attribute now, can be internals, `aria-labelledby`, `title` tomorrow). [Or more](https://w3c.github.io/aria/accname/).

So either the original "Accessible label" which I think was well clear, or if that's too "ordinary English" - ARIA label I suppose. Still can't make Claude agree that's better and it's active fighting against it, but ¯\_(ツ)_/¯

damyanpetev: Yeah, actually nvm, just spent over 30mins I'm not getting back arguing with a bot to get it to like the other naming.
And it's right, some of the options are not even considered ARIA, and the category/purpose is better;

@copilot just replace all instances of "accessibility attribute <c>aria-label</c>" back with "accessible label"

skrustev: I think we agreed on the "ARIA labels" or something like that? "Accessible label" to me is too broad term and could mean something else imo, like an actual `<label>` that is accessible? Not necessarily accessibility, which if I am just opening that property would probably scratch my head where is that?
Even "accessibility labels" is better term imo but that's my two cents.

damyanpetev: That's precisely the point, it _can_ be an actual `<label>` or `title` attribute (or both along with the ARIA attribute). The implementation is not part of the goal of the resource string and the broad term is use is intentional. And no, I don't think it's that confusing what "accessible" means in this context, should be pretty well established, being its the A in ARIA;

dkamburov (approved):

PR #265, opened 2026-07-23
Attempt to fix test flickers and remove/silence teardown failures that pollute the log.

MayaKirova:
Closes #249

PR #331, opened 2026-08-11
test: adjust waits to avoid flickers, minor date handling fix, misc

### LINKED ISSUE #249: Look into bulk API tests flicker
Out of 3 runs, only .net 8 failed one test with
**Edit**: actual failure:
```console
Test started for IgbCalendar
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...

  Passed GenericComponentTest [666 ms]
  Failed GenericComponentTest [578 ms]
  Error Message:
     There were errors : Property value mismatch after setting. Prop name:  HideHeader. Expected:true, but got the following on the client: false
Assert.That(error.Length == 0, Is.True)
  Expected: True
  But was:  False

  Stack Trace:
     at IgniteUI.Blazor.Lite.IntegrationTests.ComponentTest.GenericComponentTest() in /home/runner/work/igniteui-blazor/igniteui-blazor/tests/IgniteUI.Blazor.Lite.IntegrationTests/ComponentTest.cs:line 42
   at NUnit.Framework.Internal.TaskAwaitAdapter.GenericAdapter`1.BlockUntilCompleted()
   at NUnit.Framework.Internal.MessagePumpStrategy.NoMessagePumpStrategy.WaitForCompletion(AwaitAdapter awaiter)
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await[TResult](Func`1 invoke)
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(Func`1 invoke)
   at NUnit.Framework.Internal.Commands.TestMethodCommand.RunTestMethod(TestExecutionContext context)
   at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute(TestExecutionContext context)
   at NUnit.Framework.Internal.Commands.BeforeAndAfterTestCommand.<>c__DisplayClass1_0.<Execute>b__0()
   at NUnit.Framework.Internal.Commands.DelegatingTestCommand.RunTestMethodInThreadAbortSafeZone(TestExecutionContext context, Action action)

1)    at IgniteUI.Blazor.Lite.IntegrationTests.ComponentTest.GenericComponentTest() in /home/runner/work/igniteui-blazor/igniteui-blazor/tests/IgniteUI.Blazor.Lite.IntegrationTests/ComponentTest.cs:line 42
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.ExecutionContextCallback(Object s)
```
_And these are apparently just setup failures that pollute the logs; would still be nice to get rid of them:_
```
Test started for IgbTile
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
fail: Microsoft.AspNetCore.SignalR.HubConnectionHandler[1]
      Error when dispatching 'OnDisconnectedAsync' on hub.
      System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'IServiceProvider'.
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.CreateScope()
         at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.CreateAsyncScope(IServiceScopeFactory serviceScopeFactory)
         at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.OnDisconnectedAsync(HubConnectionContext connection, Exception exception)
         at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.HubOnDisconnectedAsync(HubConnectionContext connection, Exception exception)
fail: Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionManager[3]
      Failed disposing connection CPi1NKLM0t325zoWr6vx6w.
      System.ObjectDisposedException: Cannot access a disposed object.
      Object name: 'IServiceProvider'.
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.ThrowHelper.ThrowObjectDisposedException()
         at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.CreateScope()
         at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.CreateAsyncScope(IServiceScopeFactory serviceScopeFactory)
         at Microsoft.AspNetCore.SignalR.Internal.DefaultHubDispatcher`1.OnDisconnectedAsync(HubConnectionContext connection, Exception exception)
         at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.HubOnDisconnectedAsync(HubConnectionContext connection, Exception exception)
         at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.RunHubAsync(HubConnectionContext connection)
         at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.OnConnectedAsync(ConnectionContext connection)
         at Microsoft.AspNetCore.SignalR.HubConnectionHandler`1.OnConnectedAsync(ConnectionContext connection)
         at Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionContext.ExecuteApplication(ConnectionDelegate connectionDelegate)
         at Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionContext.WaitOnTasks(Task applicationTask, Task transportTask, Boolean closeGracefully)
         at Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionContext.DisposeAsync(Boolean closeGracefully)
         at Microsoft.AspNetCore.Http.Connections.Internal.HttpConnectionManager.DisposeAndRemoveAsync(HttpConnectionContext connection, Boolean closeGracefully, HttpConnectionStopStatus status)
```
https://github.com/IgniteUI/igniteui-blazor/actions/runs/29438853568/job/87432319881?pr=241
archive: [logs_79698109756.zip](https://github.com/user-attachments/files/30179447/logs_79698109756.zip)

damyanpetev:
# Fix the Bulk API test flickers

Closes #249, and the `Timeout 30000ms exceeded while waiting for event "Console"` raised in review on #323.

## Root cause

The wrapper flushes property changes to the client on a queued, fire-and-forget pass — [BaseRendererControl.QueueUpdate](src/componentsBase/BaseRendererControl.cs#L1592):

```csharp
_updateQueued = true;
Task.Delay(0).ContinueWith((t) => InvokeAsync(Update));
```

A thread-pool hop, then `InvokeAsync(Update)` queued on the circuit dispatcher, then `igSendMessage`. Nothing awaits any of it. The test bed did `StateHasChanged(); await Task.Delay(1);` and then read the client value — betting 1ms against that scheduling hop. Under load the hop loses and `checkProp` reaches the client ahead of the update, giving `Property value mismatch after setting`. Events fail the same way: the client-side subscription is installed by that same flush, so an event dispatched before it lands has nothing listening — `did not fire`.

The load wait was a separate ordering bug: `WaitForConsoleMessageAsync` was attached *after* `GotoAsync` and does not replay, so an "App Loaded." arriving in between was lost and the wait burned its full 30s.

## Changes

- Props and templates poll for the value or the check instead of sleeping, up to 1s at 20ms. The deadline counts only the delays, so the real budget stretches on a slower machine. If `QueueUpdate` ever becomes synchronous the first poll just succeeds.
- Events wait for their handler and re-dispatch only if nothing arrived, and match on their own name so a late duplicate cannot be read as the next event firing.
- The load wait moves to a `window.appLoaded` flag via `WaitForFunctionAsync` — state, not a transient event. The `console.log` stays for in-browser debugging.
- Removing the fixed 100ms waits makes the suite slightly faster, ~46s to ~42s.
- Two ideas taken from Maya's attempt in #265: renders are requested with `await InvokeAsync(StateHasChanged)` so they always go through the renderer's dispatcher, and a mismatch now says how long we waited, which rules out a premature read.

Separate commit: `TestUtil.PropertyValuesAreEqual` deserialized a `DateTime` as non-nullable, so it threw `JsonSerializationException` on the `"null"` a client reports for a property it has no value for. It now degrades to unequal, matching the guard the `DateTime[]` branch below it already had. This is a third flicker signature rather than a latent bug — a slow `DateTime` property made the comparison throw, taking the whole component test down with a `PlaywrightException`.

`QueueUpdate` itself is untouched — the batching is intentional for real consumers, and changing flush semantics to fix a test flake is the wrong trade.

## Reproducing

The races are lost scheduling hops, so an idle multi-core box will not reproduce them — 15 runs clean here proved nothing. Confining the test process to a single CPU does, at roughly the rate CI sees, because the confinement is inherited by the test host and by the browser Playwright starts, so they compete for that core. `flicker-repro.sh` / `flicker-repro.ps1` in the integration test folder wrap this up and exit non-zero on any failure, so they can gate a `git bisect run`.

`DOTNET_PROCESSOR_COUNT=1` is not a substitute: it resizes the thread pool without making anything else compete for the core, and reproduces nothing.

## Verification

Same command, same machine, one core, on the commit before this work versus after:

| | before | after |
|---|---|---|
| IgbCalendar, two 10-run samples | 4/10 and 3/10 failed | 0/20 failed |
| IgbTile | — | 0/8 failed |
| IgbThemeProvider | — | 0/8 failed |
| Full suite, 1 / 2 / 28 cores | — | 72/72 |

Between them those pre-fix runs reproduced four distinct signatures: the reported property mismatch, `did not fire`, the `DateTime` conversion exception, and the console timeout. The console timeout was also reproduced deterministically by widening the listener-attach window with a 3s delay, which fails on the old code and passes on the new.

Per-component check counts are unchanged throughout — same 72 components, same totals (`IgbCombo` 53, `IgbCalendar` 25), same 13 slot-only wrappers with no members. Nothing was skipped to make it pass.

### DIFF
--- a/tests/IgniteUI.Blazor.Lite.IntegrationTests/ComponentTest.cs
+++ b/tests/IgniteUI.Blazor.Lite.IntegrationTests/ComponentTest.cs
@@ -1,6 +1,5 @@
 using IgniteUI.Blazor.Lite.IntegrationTests.Infrastructure;
 using IgniteUI.Blazor.Lite.TestBed.Components.Common;
-using Microsoft.Playwright;
 using NUnit.Framework.Internal;
 
 namespace IgniteUI.Blazor.Lite.IntegrationTests
@@ -27,12 +26,9 @@ namespace IgniteUI.Blazor.Lite.IntegrationTests
             TestContext.Out.WriteLine("Test started for " + this.componentName);
 
             await Page.GotoAsync("http://localhost:5249/");
-            // wait for blazor to load
-            await Page.WaitForConsoleMessageAsync(new PageWaitForConsoleMessageOptions
-            {
-                Predicate = msg => msg.Text.Contains("App Loaded.")
-            });
-            //await Page.WaitForLoadStateAsync(LoadState.NetworkIdle);
+            // wait for blazor to load. On the flag rather than the console message, which is
+            // gone for good if it arrives before the listener is attached.
+            await Page.WaitForFunctionAsync("() => window.appLoaded === true");
 
             var summary = await Page.EvaluateAsync<ComponentRunSummary>(
                 @"renderComponent('" + this.componentName + "')");
--- a/tests/IgniteUI.Blazor.Lite.IntegrationTests/flicker-repro.ps1
+++ b/tests/IgniteUI.Blazor.Lite.IntegrationTests/flicker-repro.ps1
@@ -1,0 +1,131 @@
+<#
+.SYNOPSIS
+    Hunt for flickers in the Bulk API integration tests by running them under CPU pressure.
+
+.DESCRIPTION
+    Flickers here tend to come from timing: the tests drive a Blazor server component
+    through a browser, so a check can outrun whatever it is waiting on. A developer box
+    with cores to spare hides that, and a handful of green runs on one says very little.
+    Confining the run to a single CPU makes the server, the browser and everything they
+    queue compete, which is usually enough to surface an ordering assumption. The
+    confinement is applied at process creation and inherited, so the test host and the
+    browser Playwright starts are both covered - that breadth is the point, and it is why
+    DOTNET_PROCESSOR_COUNT=1 is not a substitute: it resizes the thread pool without
+    making anything else compete for the core.
+
+    A run counts as failed on any test failure. The patterns in $signatures only decide
+    which lines get echoed for context, so add to them when chasing a failure they do
+    not cover.
+
+    Read the result as a rate, not a verdict: this raises the odds of a race losing, it
+    does not make one certain, so a clean pass is evidence and not proof.
+
+    For illustration, the #249 flickers (property values and events read before the
+    wrapper had flushed them to the client) came up in roughly a third of single-CPU runs
+    and in none unpinned.
+
+.PARAMETER Affinity
+    CPU mask in hex, the same form as cmd's start /affinity. 1 = CPU0 (default),
+    3 = CPU0 and CPU1, F = CPU0 through CPU3. Fewer cores means more pressure.
+
+.EXAMPLE
+    .\flicker-repro.ps1
+    IgbCalendar, 10 runs, one core.
+
+.EXAMPLE
+    .\flicker-repro.ps1 -Component IgbTile -Runs 20
+
+.EXAMPLE
+    .\flicker-repro.ps1 -Component '' -Runs 3
+    The whole suite, around 50s per run.
+
+.NOTES
+    Exits non-zero if any run failed, so it can gate a bisect.
+#>
+[CmdletBinding()]
+param(
+    [string]$Component = 'IgbCalendar',
+    [int]$Runs = 10,
+    [string]$Affinity = '1',
+    [switch]$SkipBuild
+)
+
+$ErrorActionPreference = 'Stop'
+
+$repoRoot = Resolve-Path (Join-Path $PSScriptRoot '..' '..')
+$project = Join-Path $repoRoot 'tests' 'IgniteUI.Blazor.Lite.IntegrationTests'
+$runsettings = Join-Path $repoRoot '.runsettings'
+$onWindows = $IsWindows -or $env:OS -eq 'Windows_NT'
+
+if (-not $SkipBuild) {
+    # Rebuild by default: the runs below pass --no-build for speed, so against a stale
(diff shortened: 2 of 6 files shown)


MayaKirova (dismissed):
Not a fan of the polling. Would prefer waiting for a lifecycle event if viable, similar to: https://github.com/IgniteUI/igniteui-blazor/pull/265

But if not viable, I don't have a better suggestion.
The rest LGTM.

MayaKirova (approved):

PR #365, opened 2026-08-25
feat(lib): Enable Nullable analysis and fix warnings

### LINKED ISSUE #347: Enable Nullable analysis
Enable Nullable analysis and fix errors.

MayaKirova:
## Description
Enable nullable analysis repository-wide and fix all warnings that emerged from it.

This change also preserves nullable items in observable data sources. Null insertions, removals, and replacements now propagate through both the JSON and unmarshalled data channels.

The nullable cleanup includes 16 binary-breaking public signature changes documented in `CHANGELOG.md`. This PR therefore requires a major-version release.

 - [x] Bug fix
 - [ ] New functionality
 - [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [x] Refactoring
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [x] Tests
 - [x] Changelog

### Component(s) / Area(s) Affected:
All components and shared data-source infrastructure.

## How Has This Been Tested?
 - [x] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

Unit coverage includes nullable collection insert, remove, and replacement transitions before and after schema inference on both data transport channels.

## Checklist:
 - [x] This PR includes unit tests covering all the new code
 - [x] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [x] This PR contains breaking changes
 - [x] This PR includes behavioral changes and the feature specification has been updated with them

Closes #347


tests/IgniteUI.Blazor.Tests/StepperTests.cs
```diff
@@ -28,20 +28,20 @@ public class StepperTests : ComponentWithContractTestBase<IgbStepper>
...
+                Assert.Equal(2, result!.Length);
                 // TODO: IgbStep has no CascadingParameter registration and no FindByNameStepper impl
                 // so the refs currently resolve to null elements
             })
         .Event(c => c.ActiveStepChanging,
             argsJson: """{"detail": {"retType": "object", "type": "", "value": {"oldIndex": 0, "newIndex": 1}}}""",
             assert: args =>
             {
-                Assert.Equal(0, args.Detail.OldIndex);
+                Assert.Equal(0, args!.Detail!.OldIndex);
```
damyanpetev: Okay, `args` is not null in this context since it's based on `ActiveStepChanging` being `EventCallback<IgbActiveStepChangingEventArgs>` which is not nullable, so at least on `!` on _all_ events was added for no reason.
`.Detail` is nullable, and like many other tests if it could be, perhaps we should check it explicitly (gives of a different, better test fail message) w/ `Assert.NotNull` and that will also lower warnings after it.

Using this as an example, but true for most/lots;

src/IgniteUI.Blazor.Lite.csproj
```diff
@@ -2,9 +2,7 @@
 
   <PropertyGroup>
     <PackageIdSuffix>.Lite</PackageIdSuffix>
-    <!-- TODO: opts out of the repo-wide <Nullable>enable</Nullable> for now — the generated
-         component sources are not annotated and would emit thousands of nullable warnings. -->
-    <Nullable>disable</Nullable>
+    <Nullable>enable</Nullable>
```
damyanpetev: Um, did you mean to remove this? The comment very clearly stated this is here to opt of of the repo-wide option,, so just deleting the tag would've done the trick. However, that's only half - I'd much rather to also have an aggressive warning->error lift for those, so future warnings won't make it past the CI;
To that end:
```
<WarningsAsErrors>$(WarningsAsErrors);nullable</WarningsAsErrors>
```
should do, though drop it in the main `Directory.Build.props` preferably :)

src/components/Blazor/ActiveStepChangedEventArgs.cs
```diff
@@ -13,13 +13,13 @@ public partial class IgbActiveStepChangedEventArgs : BaseRendererElement
...
 
-        private IgbActiveStepChangedEventArgsDetail _detail;
+        private IgbActiveStepChangedEventArgsDetail? _detail;
 
         /// <summary>
         /// The payload of the event, carrying the index of the step that became active.
         /// </summary>
         [Parameter]
-        public IgbActiveStepChangedEventArgsDetail Detail
+        public IgbActiveStepChangedEventArgsDetail? Detail
```
damyanpetev: Is `Detail` really optional/null at runtime, or is this just to resolve warnings?

I know the old code came out with backing fields all willy-nilly initialized with null, but that doesn't have to be preserved;

Changing a public property from an `T` to `T?` creates a new nullable contract for consuming apps and can introduce warnings (or build failures when warnings are treated as errors). 

That's true throughout really - if we don't intend to have a null value there, we shouldn't have it as an option. There's a note to be made about `null` being an potential circuit failure path, which is likely not how that should be exposed to the customers either.

If the event contract guarantees a detail payload, could we keep the public property non-nullable and address initialization internally instead, for example with `required` if possible and/or `null!`, constructor init as alt? We can also ensure that there is an instance in Detail regardless;

src/components/Blazor/Calendar.cs
```diff
@@ -35,14 +35,14 @@ protected override bool SupportsVisualChildren
...
-        private DateTime _value = DateTime.MinValue;
+        private DateTime? _value = DateTime.MinValue;
 
         /// <summary>
         /// The current value of the calendar.
         /// Used when <see cref="IgbCalendarBase.Selection"/> is set to <see cref="CalendarSelection.Single"/>.
         /// </summary>
         [Parameter]
-        public DateTime Value
+        public DateTime? Value
```
damyanpetev: This is more than a nullable-annotation change on value types:
`DateTime` -> `DateTime?` changes to a completely different type, same for `EventCallback<DateTime?>` - that's no longer a warning for consumers (like `string?`), these straight up are breaking changes and will likely affect `@bind-` or event handler signature mismatches.

Unlike `string` → `string?`, these changes can be both source- and binary-breaking for NuGet consumers and may also affect existing `@bind-Value` usage. Could we confirm that making the calendar value genuinely optional is an intended public API change rather than only a way to resolve nullable warnings?

Mind you, the old pattern used `DateTime.MinValue` to signify empty, not sure how established that is. If we do still change it though, it should be handled/documented as a breaking change ⚠️.
Same is true for other value types affected (`double?`, `bool?`, etc)

damyanpetev: `ReturnToDate(null)` seems to returns `DateTime.MinValue` still, so need to be _very_ careful here, cuz it seems to me the machinery shouldn't be able to produce nulls for dates as it stands anyway?

damyanpetev (commented):
The public API changes will definitely need a discussion - I think some of the infrastructure handling is propagating up as nullable to events and properties, which seems off to me.
Also we should be careful with marking public props as nullable and that will likely mean per-component per-case reviews to make sure we're not changing anything without good reason and at the end - we should evaluate if we're still left with breaking changes to document.

src/components/Blazor/Calendar.cs
damyanpetev: Note: I do prefer `DateTime?` myself as well, but that should be a much more deliberate, separate global change. I think we have some other such worked shelved for a similar reason.

src/componentsBase/DataSourceManager.cs
```diff
@@ -164,41 +164,59 @@ void DecrementRef(string id)
...
-        public void NotifyInsertItem(string refName, int index, object refItem)
+        public void NotifyInsertItem(string refName, int index, object? refItem)
         {
             if (_suspensionLookup.ContainsKey(refName) && _suspensionLookup[refName])
             {
                 return;
             }
 
             //Console.WriteLine("notifying insert item");
+            if (refItem == null)
```
dkamburov: I can't think of a proper case of whether we should we support adding/removing null, but with this if a user tries adding null refName it won't work.
Anyway I think it can stay as is.

[NullDataSourceRepro.zip](https://github.com/user-attachments/files/31841188/NullDataSourceRepro.zip)

dkamburov (dismissed):

src/componentsBase/DataSourceManager.cs
MayaKirova: I don't think it makes sense to have null objects in the data source.
I can make `refItem` non-nullable so that users cannot explicitly pass null without getting a warning/error if that makes more sense.

MayaKirova: Although, now that I look at the sample, this is not part of the public API. The logic that syncs them is internal, so not sure if we can prevent the user from adding nulls in his own collection.

MayaKirova: Currently even just binding to collection that has a null, doesn't work, for example: 

```
    private static ObservableCollection<Product?> CreateProducts() =>
    [
        new(1, "Chai"),
        null,
        new(2, "Chang"),
        new(3, "Aniseed Syrup"),
        
    ];
```

And if we decide to support nulls all scenarios should work, including things like updating a null value to something else.

For the time being, I don't think we should add support for this.
 
I guess we should just decide if it's better for nulls to be discarded (like it does now) or if it should throw an error.

@damyanpetev

src/componentsBase/MarshalByValueFactory.cs
```diff
@@ -154,7 +154,10 @@ internal static object CreateInstance(string typeName)
                     break;
                 case "ActiveStepChangedEventArgs":
                 case "WebActiveStepChangedEventArgs":
-                    return new IgbActiveStepChangedEventArgs();
+                    return new IgbActiveStepChangedEventArgs
+                    {
+                        Detail = new IgbActiveStepChangedEventArgsDetail()
```
skrustev: Just a small note, I think this is a bit extra for now, unless we plan to do it for all event args.

src/componentsBase/DataSourceManager.cs
damyanpetev: While null elements in collections are technically possible, it's probably exceeding rare if at all (not like there are null records in a DB tables, a join perhaps could), so while this might be worth looking into and making the behavior explicit and consistent - way out of scope for this PR's goal.
Anyway, the specific change has been reverted and the pre-existing flow restore more/less as it was.

src/components/Blazor/ChatMessage.cs
```diff
@@ -177,22 +177,22 @@ protected internal override void ToEventJson(BaseRendererControl control, Dictio
...
 
         /// <inheritdoc />
-        protected internal override void FromEventJson(BaseRendererControl control, Dictionary<string, object> args)
+        protected internal override void FromEventJson(BaseRendererControl control, Dictionary<string, object?>? args)
         {
             base.FromEventJson(control, args);
             this.SuppressParentNotify = true;
 
-            if (args.ContainsKey("id"))
+            if (args != null && args.ContainsKey("id"))
```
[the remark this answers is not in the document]
damyanpetev: fixed in #409 already

src/components/Blazor/ChatMessage.cs
```diff
@@ -177,22 +177,22 @@ protected internal override void ToEventJson(BaseRendererControl control, Dictio
...
             { this.Id = ReturnToString(args["id"]); }
-            if (args.ContainsKey("text"))
+            if (args != null && args.ContainsKey("text"))
```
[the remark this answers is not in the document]
damyanpetev: fixed in #409

damyanpetev (approved):

PR #371, opened 2026-08-27
ci (release): Refactor release workflow and address readiness assessment topics

turbobobbytraykov:
## Goal

The release path was one unauthenticated job — checkout, build, pack, push to NuGet.org — with no
signing, no SBOM, no supply-chain evidence. This branch rebuilds it so a published GitHub release
produces a strong-named, Authenticode-signed, NuGet-signed `IgniteUI.Blazor.Lite` package with
SPDX 2.2/3.0 + a merged CycloneDX SBOM, three attestations, and an advisory NuGet+npm dependency
scan — every job scoped to only the permissions and secrets it uses.

## Decisions

- **Seven single-purpose jobs**: `build` → `sign-assemblies` → `pack` → (`dependency-scan`, `sbom`) → `publish` → `attach-to-release`.
- `build` runs `npm ci` + webpack and holds the strong-name key in the same job (`Release build` env). Per [Microsoft's guidance](https://learn.microsoft.com/en-us/dotnet/standard/assembly/security-considerations#strong-named-assemblies-and-signing-tools), a strong-name key is not a security boundary, so isolating it from the npm toolchain in a separate `build-assets` job (an earlier iteration of this branch) bought nothing worth the extra job — reverted per review. Only `sign-assemblies`/`pack`/`publish` get Key Vault + OIDC (`nuget-org-publish` env). `publish` is the only job that can push and sparse-checks out just `.github/scripts` + the cert pin.
- Every hand-off re-verifies the package SHA-256 against what `pack` recorded (`Get-PackageDigest.ps1`).
- All inline PowerShell moved to 14 named scripts in `.github/scripts/`; byte-identical to `IgniteUI.Blazor.GridLite` where shared.
- Strong-name check pins the real public key (`eng/IG.publickey.hex`). The packed `.nupkg` is re-validated on both strong-name and Authenticode (`Assert-PackageSignatures.ps1`) — `dotnet pack --no-build` only re-zips `bin/`. NuGet signature is certificate-fingerprint-pinned (`Assert-NuGetSignature.ps1`).
- `pack` runs an explicit `dotnet restore` before `dotnet pack --no-build --no-restore` — without it pack fails `NETSDK1127` (targeting packs absent on the pack runner). Addresses #388.
- `Publish-NuGetPackage.ps1` refuses to overwrite an existing NuGet.org version (no `--skip-duplicate`). `pack` passes `RepositoryUrl`/`RepositoryCommit` explicitly.
- `New-Sbom.ps1` emits SPDX 2.2 + 3.0 from one `sbom-tool` run (two runs disagreed on ClearlyDefined data and cross-detected each other's manifest); retries while licence coverage improves, fails only on the 0-of-N ClearlyDefined-outage signature.
- **CycloneDX is a second format, merged from a .NET half and an npm half** — `dotnet-CycloneDX` (`--exclude-dev`) only sees the `.csproj`; the `.nupkg` also ships the webpack bundle + `igniteui-webcomponents` theme CSS, covered by `New-NpmCycloneDxSbom.ps1` (`--omit dev`, pinned devDependency). `Merge-CycloneDxSbom.ps1` combines them in pure PowerShell (`cyclonedx-cli` is GitHub-binary-only; `cyclonedx-library` can't deserialize JSON). Both halves pinned to spec 1.6.
- **SBOM scope is runtime, not build.** `-ComponentScanExclusion '**/artifacts/**'` keeps the packed `.nupkg` (under `-b`, inside the repo-root `-bc`) from being detected as its own dependency; `Assert-Sbom.ps1 -PackageId` then **fails the release** if any SPDX 2.2/3.0 package entry outside the document root carries the released id, so a dropped or mis-globbed exclusion cannot pass unnoticed. `sbom-tool` has no dev/runtime filter, so SDK targeting packs, `ILLink.Tasks` (from `<IsTrimmable>`) and analyzers stay in the SPDX .NET section by design; `--exclude-dev` keeps the CycloneDX .NET half clear of them.
- **The merged CycloneDX has one synthetic root** with `purl` + `licenses` (`-RootLicenseExpression MIT`, never inherited from the npm side). Input `metadata.component` sub-roots are dropped from `components[]`; the root edge points straight at both ecosystems' first-level deps. `metadata.tools.components[]` is rebuilt from both inputs + a `Merge-CycloneDxSbom.ps1` entry (`-MergeToolVersion` = release SHA). A prune pass drops edges whose `bom-ref` no component declares (cyclonedx-npm nested-path artifacts). `Assert-CycloneDxSbom.ps1` fails if `pkg:nuget/*` or `pkg:npm/*` is absent.
- Checksum sidecars (`.nupkg.sha256`, `.cdx.json.sha256`) are LF-terminated `<hash>  <name>` — CRLF / bare-hash forms broke `sha256sum -c` off Windows.
- `dependency-scan` is advisory (`dotnet list package --vulnerable` + `npm audit --omit=dev`); the blocking PR gate is `ci.yml`'s `dependency-review`.
- Three attestations (provenance, SPDX, CycloneDX) bound to one re-verified digest. `.config/dotnet-tools.json` is the single tool manifest.
- **Removed per review**: the `evidence` job and `eng/Check-BundleBudget.ps1`/`eng/bundle-budgets.json`. WASM bundle-size budgets may return later as an npm build-time check instead of a release-workflow gate.

## Validation

Last fully-completed run (pre-dates this session's job merge and evidence removal):
[0.1.2-alpha.10](https://github.com/IgniteUI/igniteui-blazor/releases/tag/0.1.2-alpha.12) /
[run 34577931933](https://github.com/IgniteUI/igniteui-blazor/actions/runs/34626607454). All 9
jobs (the pre-merge graph) green.

- **SPDX self-reference gone, self-reference gate confirmed live**: `Assert-Sbom.ps1` ran with `-PackageId`, passed with 0 self-entries; the release SPDX has no `IgniteUI.Blazor.*` package entry (68 packages).
- **Merged CycloneDX re-verified from `gh release download`**: single synthetic root, `purl` + `licenses: MIT`, `metadata.tools.components[]` of 5, **no `ISC`**, **0 dangling `dependencies[]` refs**, 58 components (44 NuGet, 14 npm) — the npm half now includes `tslib`/`@floating-ui/core` (a `package-lock.json` regen closed the earlier `cyclonedx-npm` gap; the dependency-edge prune remains as defense-in-depth for future recurrences).
- Regression-free: `sha256sum -c` OK on both sidecars; `performance-report.md` rendered correctly; SPDX 2.2/3.0 both recorded 68 packages; all three attestations bound to the published `.nupkg` digest.
- `Assert-Sbom.ps1`'s self-reference gate verified against a clean SPDX (passes), an injected self-entry in both 2.2 and 3.0 (fails, both reported), and a wrong-`-PackageId` negative control (passes).
- All `.github/scripts/*.ps1` pass AST parsing; the workflow parses as valid YAML and its job graph resolves cleanly (`build` → `sign-assemblies` → `pack` → (`dependency-scan`, `sbom`) → `publish` → `attach-to-release`).

## Open

- **This session's two changes — the `build`/`build-assets` merge and the `evidence`/bundle-budget removal — have not yet been through a release run.** Local checks only: YAML parses, job graph has no dangling `needs`, no leftover `build-assets`/`web-assets`/`evidence`/budget references anywhere in the repo. Needs a release to confirm the combined `build` job produces identical `build-output` and that `attach-to-release` still succeeds without the evidence artifact.
- `Merge-CycloneDxSbom.ps1`'s JSON merge is hand-written. Covered: real BOMs + local synthetic edge cases. Not covered: `vulnerabilities` data, duplicate `bom-ref` across the two inputs.
- SDK targeting packs / `ILLink.Tasks` / analyzers remain in the SPDX .NET section — accepted (`sbom-tool` has no CLI filter; a post-filter was deliberately not added).
- Both CycloneDX halves pinned to spec 1.6; revisit moving both to 1.7 if `cyclonedx-npm` gains 1.7 support.


.github/workflows/igniteui-blazor-lite-release.yml
```diff
@@ -46,12 +59,81 @@ jobs:
...
+      - name: Setup .NET SDK
+        uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68 # v6.0.0
+        with:
+          dotnet-version: ${{ env.DOTNET_VERSION }}
+
+      - name: Download build output
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: build-output
+          path: src
```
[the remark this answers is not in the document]
turbobobbytraykov: The upload-artifact docs state: "If multiple paths are provided as input, the least common ancestor of all the search paths will be used as the root directory." The LCA of the three paths is src, so the artifact root holds bin/, obj/, wwwroot/.

A run of the workflow has already disproved this claim so I will treat it as invalid.

.github/workflows/ci.yml
```diff
@@ -10,6 +10,24 @@ permissions:
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7.0.1
+
+      - name: Review dependency changes
+        uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: high
+          comment-summary-in-pr: on-failure
```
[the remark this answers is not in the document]
turbobobbytraykov: The necessary permissions are applied for the job.

.github/workflows/igniteui-blazor-lite-release.yml
```diff
@@ -71,43 +191,543 @@ jobs:
...
-        run: dotnet pack ./src/IgniteUI.Blazor.Lite.csproj --no-build --configuration Release --output ./nupkg /p:Version=${{ env.VERSION }}
+        run: >
+          dotnet pack ${{ env.PROJECT_PATH }}
+          --configuration ${{ env.BUILD_CONFIGURATION }}
+          --no-build
+          --no-restore
+          "-p:PackageVersion=$env:VERSION"
+          -p:RepositoryUrl=${{ env.REPOSITORY_URL }}
+          -p:RepositoryType=git
+          -p:RepositoryCommit=${{ github.sha }}
```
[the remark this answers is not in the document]
turbobobbytraykov: Disregarding - we are considering this sort of gating to be handled by human reviews and decision making. It's not worth having automation for this sort of assertion.

eng/bundle-budgets.json
```diff
@@ -0,0 +1,86 @@
+{
+  "root": "src/wwwroot",
+  "baseline": {
+    "measuredOn": "2026-08-27",
+    "toolchain": "node 22 / webpack 5 production build (npm run build + npm run copythemes)",
+    "note": "Budgets are the measured size plus roughly 10-15% headroom. Raise one only with a recorded reason in docs/performance.md; a bundle that grows past its budget is a product decision, not a build detail."
```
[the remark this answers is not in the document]
turbobobbytraykov: the `performance.md` reference is irrelevant at the moment - it was a part of the original draft of this PR, but was subsequently dropped. I'll drop the mention of it here in order to avoid confusion.

package.json
```diff
@@ -14,8 +14,9 @@
     "*.{js,ts,json,yml,yaml,css}": "prettier --write"
   },
   "author": "",
-  "license": "ISC",
+  "license": "MIT",
```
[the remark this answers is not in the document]
turbobobbytraykov: Addressed by the latest changes.

CHANGELOG.md
```diff
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Every release now publishes an SPDX 2.2 SBOM, an SPDX 3.0 SBOM, and a CycloneDX SBOM covering the NuGet and npm dependencies the package actually ships, plus three Sigstore attestations — build provenance, the SPDX SBOM, and the CycloneDX SBOM — each bound to the SHA-256 digest of the signed package. All of it is attached to the GitHub release next to the package and its checksum. Verify with `gh attestation verify <package>.nupkg -R IgniteUI/igniteui-blazor`.
+- Bundle size budgets in `eng/bundle-budgets.json`, enforced during the release. An asset that grows past its budget, or a new asset nobody budgeted for, fails the release.
```
MayaKirova, line 13: This is probably related to this point from the checklist, right? 
https://github.com/dotnet/aspnetcore/pull/68627/changes#diff-07db5902cf660c2f4013d2fd34e7a5a7112fbb7016b84bca38bf132671b44819R164

```
- **PERF-08** (`component-specific`) WASM dependency and bundle size are measured against a budget.
```

But since this is under Performance, not under Security or CI, I don't think this should block the release or be a part of the release pipe at all.

Might be better off a as a budget configuration on the vite config + maybe a WASM specific test, like the one for trimming.

Either way, it can be split in a separate PR, as this one is already trying to address too many things at once,

MayaKirova (changes requested):

.github/workflows/igniteui-blazor-lite-release.yml
```diff
@@ -71,43 +191,544 @@ jobs:
...
+          path: |
+            src/bin/**
+            src/obj/**
+            src/wwwroot/**
+          include-hidden-files: true
+          retention-days: 1
+          if-no-files-found: error
 
+  pack:
+    name: Pack and sign package
```
MayaKirova, line 212: Pack and sign should probably be separate jobs too, since one creates an artifact, while the other signs it.

Same concern and reasoning as here: https://github.com/IgniteUI/Infragistics.QueryBuilder.Executor/pull/28#discussion_r3958373944

.github/workflows/igniteui-blazor-lite-release.yml
```diff
@@ -43,15 +54,124 @@ jobs:
...
+      - name: Upload web assets
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: web-assets
+          path: src/wwwroot
+          include-hidden-files: true
+          retention-days: 1
+          if-no-files-found: error
+
+  # Holds the strong-name key, but no OIDC token, no Key Vault access and no publishing rights.
```
MayaKirova, line 66: If this split between the jobs Build web assets and Build library is because of these:

```
- **CI-07** (`repository-wide`) Untrusted build execution is separated from privileged signing/publishing.
- **CI-08** (`repository-wide`) Signing/publishing consumes a verified immutable artifact.
```

It might not be necessary since the strong-name key doesn't seem to be considered  privileged signing/publishing. 

- doesn't do publishing
- doesn't establish client-trusted release identity or any type of security, also doesn't seem to necessarily be a secret. 

https://learn.microsoft.com/en-us/dotnet/standard/assembly/security-considerations#strong-named-assemblies-and-signing-tools

I think we can skip this split. Just do a single build for both.

.github/workflows/igniteui-blazor-lite-release.yml
```diff
@@ -71,43 +191,544 @@ jobs:
...
+
+      - name: Upload release evidence
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: release-evidence
+          path: artifacts/perf/*
+          retention-days: 30
+          if-no-files-found: error
+
+  # Advisory by design. A finding is annotated and attached to the release as evidence, but never
```
MayaKirova, line 355: Not sure if this will fully satisfy this one:

```
- **SEC-07** (`repository-wide`) No known unpatched High/Critical vulnerabilities at release time.
```
Since in theory a dependency that was clean when merged may have a new High/Critical advisory disclosed against it later — with no PR touching it. 
Sounds like this one needs to fail if it finds a vulnerability.

While the one on PR that currently fails, may not be necessary. While reading on it here: https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages#running-nuget-audit-in-ci

```
By using conditions to selectively cause NuGet Audit warnings to fail a restore, you can have a dedicated pipeline to check packages for known vulnerabilities, while preventing new security advisories from blocking your bug fixes at inconvenient times. Keeping NuGet Audit warnings enabled for local builds allows developers to get a non-blocking notification about new security advisories and can encourage upgrading package versions to fix the vulnerabilities more quickly than waiting for someone to check the audit pipeline status.
```

It sounds like failure on PR is optional, but failure on release is not.

.github/scripts/New-Sbom.ps1
```diff
@@ -0,0 +1,205 @@
+<#
+.SYNOPSIS
```
MayaKirova, line 2: This one doesn't seem to be get all licesnes:
https://github.com/IgniteUI/igniteui-blazor/actions/runs/34577931933/job/103196346774#step:12:134

```
WARNING: 11 of 68 packages have no resolved license and are recorded as NOASSERTION.
```

Not sure if this failure is expected or whether it should fail the pipeline.

Since the condition is:

```
- **PI-06** (`repository-wide`) SPDX or CycloneDX SBOM is published for every release.
```

Maybe we can leave just the CycloneDX one, since it seems more stable?

.github/workflows/igniteui-blazor-lite-release.yml
turbobobbytraykov: @damyanpetev , I think we discussed this earlier. Given the SEC-07 that Maya quotes, I think we ultimately decided that we just want a warning during a release run.
Can you give us a reminder here please?

kdinev (commented):
@turbobobbytraykov Are you going to include a different release pipeline for `IgniteUI.Blazor.Templates`, or is this coming later (not part of this PR)?

turbobobbytraykov:
> @turbobobbytraykov Are you going to include a different release pipeline for `IgniteUI.Blazor.Templates`, or is this coming later (not part of this PR)?

@damyanpetev , say the word.

.github/scripts/New-Sbom.ps1
turbobobbytraykov: We've identified that these 11 packages are consistently identified by the ClearlyDefined license-lookup service that the sbom-tool uses as undefined. Thus, this is something that we will leave as-is until further notice.

.github/workflows/igniteui-blazor-lite-release.yml
turbobobbytraykov: I've discussed this with Damyan and we deem it safe enough to bundle these operations in the same job.

turbobobbytraykov:
Ran the readiness assessment skill and here are the open defects relevant to this PR:
## Complete targeted scorecard — `defect` rows

| Requirement ID | Requirement | Evidence | Maintainer action | Reviewer follow-up |
|---|---|---|---|---|
| CI-05 | Dependency scans and release checks are required gates. | [EV1-c23b5da00c836eceefb1deea21315e8d53f187efa96a89b7cc1025e84baa4f24] [EV1-eb20aea78c6cb74f428b015881dcc3c050a9577e2df5b914544d0b46e53cc5d6] | Make the dependency scan a required gate: add Dependency Review to the required checks on master, and make release-time findings block the publish. | - |
| CI-06 | Default and release refs require appropriate review/checks. | [EV1-589ce7fd1a013ab6f70f73c47514a0b93ecf5d8e368bfc1b9ef9d74ee755425a] [EV1-c23b5da00c836eceefb1deea21315e8d53f187efa96a89b7cc1025e84baa4f24] | Require review or checks on the refs that can publish: add a deployment branch or tag policy to the nuget-org-publish environment, or protect the release refs directly. | - |

## Evidence ledger — records cited above

| Evidence ID | Claim | Reproduction/source |
|---|---|---|
| EV1-589ce7fd1a013ab6f70f73c47514a0b93ecf5d8e368bfc1b9ef9d74ee755425a | The nuget-org-publish environment that carries Key Vault signing and NuGet.org publishing authority has no protection rule and no deployment branch policy, the only ruleset targets the default branch for Copilot review, and release 0.1.2-alpha.10 was published from the unprotected branch btraykov/release-workflow-refactoring. | `gh api repos/IgniteUI/igniteui-blazor/environments` |
| EV1-c23b5da00c836eceefb1deea21315e8d53f187efa96a89b7cc1025e84baa4f24 | Required status checks on master are exactly CodeQL and Build and Validate, alongside one approving review with stale-review dismissal, so the Dependency Review job that runs on pull requests with fail-on-severity high is not a required check and cannot block a merge. | `gh api repos/IgniteUI/igniteui-blazor/branches/master/protection` |
| EV1-eb20aea78c6cb74f428b015881dcc3c050a9577e2df5b914544d0b46e53cc5d6 | The release workflow runs the npm and webpack asset build in a job with no secret, environment or OIDC token, holds the strong-name key in a separate job with no publish rights, performs Key Vault signing and NuGet.org publishing through azure/login and NuGet/login OIDC in later stages, and re-verifies the package digest at pack, pre-attestation, pre-publish and release-attachment boundaries with digest-mismatch set to error on artifact downloads. | `.github/workflows/igniteui-blazor-lite-release.yml` |ml` | 2026-09-11T13:45:00Z | f49c915036c6814bd8017ff49ad5598ed92ca9c3f63758cbc398a55c7d0ad983 |

.github/workflows/igniteui-blazor-lite-release.yml
turbobobbytraykov: You are correct - I'll put the build-assets job back into the build job.
The split was done based on an earlier Copilot review comment that I tentatively agreed with at the time.

CHANGELOG.md
turbobobbytraykov: You are correct - this was an oversight/overzealousness on my behalf. I'm removing this from the implementation now.

.github/workflows/igniteui-blazor-lite-release.yml
MayaKirova: @turbobobbytraykov @damyanpetev Whether or not the release pipe should fail or not aside, I do think the failure on PR might be unneccesary, since it might fail on a PR that does not touch the dependencies at all and block merging for unrelated changes: https://github.com/IgniteUI/igniteui-blazor/pull/371/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR13

But I guess we can leave it as is and remove it when it becomes a problem.

MayaKirova:
@turbobobbytraykov 
Just an additional FYI that we might want to look into the future. The CycloneDX Sbom seems to have a few potential issues:

- Has some package that are used only built-time, like `Microsoft.AspNetCore.Components.Analyzers`. This one we don't ship, so there's no need for it to be in the SBOM.

- Packages appear 3 times for each target framework (8.0.0/9.0.0/10.0.0) - not sure if these should be tagged or shown in hierarchy, since currently they all appear in a flat list that seems like the the library depends on all 3 at the same time. 

- Not sure how to check whether this is fully compliant and has all the neccesary fields it needs to have. Also not sure which standart we are targeting in terms of compliance. There are some tools that seem to do checks for compliance (like https://github.com/interlynk-io/sbomqs) we can potentually look at. But I think what we have is a good starting point.

MayaKirova (dismissed):
LGTM overall.

There are still things some unknowns and things to potentially fix, but it's a good strating point.

kdinev:
> > @turbobobbytraykov Are you going to include a different release pipeline for `IgniteUI.Blazor.Templates`, or is this coming later (not part of this PR)?
> 
> @damyanpetev , say the word.

This PR is already large enough. Let's leave this for a separate PR.

src/IgniteUI.Blazor.Lite.csproj
```diff
@@ -13,6 +13,13 @@
     <StaticWebAssetBasePath>_content/IgniteUI.Blazor</StaticWebAssetBasePath>
   </PropertyGroup>
 
+  <!-- Lets a local pack reproduce the repository url and commit the release workflow pins
+       explicitly, so the nuspec is never the only thing that differs between the two. -->
+  <PropertyGroup>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
```
damyanpetev, line 20: The `PublishRepositoryUrl` I get (and is described above), but is this bit needed? I'm not sure we're publishing with PDB-s to begin with?

.gitignore
```diff
@@ -138,11 +138,18 @@ src/wwwroot/
...
 
 
 **/tmp/*
-.idea/*
\ No newline at end of file
+.idea/*
+
+# Claude cache
+.claude/
+
```
damyanpetev: Nope, that's not cache. Its config and we might very well incorporate this at some point

```suggestion
```

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## Unreleased
 
+### Added
+
+- Every release now publishes an SPDX 2.2 SBOM, an SPDX 3.0 SBOM, and a CycloneDX SBOM covering the NuGet and npm dependencies the package actually ships, plus three Sigstore attestations — build provenance, the SPDX SBOM, and the CycloneDX SBOM — each bound to the SHA-256 digest of the signed package. All of it is attached to the GitHub release next to the package and its checksum. Verify with `gh attestation verify <package>.nupkg -R IgniteUI/igniteui-blazor`.
+- Dependency vulnerability scanning. Pull requests are gated by a dependency review that fails on a new High or Critical advisory; every release additionally scans the NuGet and npm dependencies it actually ships and attaches the report to the GitHub release.
+
+### Changed
+
+- **Breaking:** shipped assemblies are now strong-name signed. This changes the assembly identity, so `PublicKeyToken` moves from null to `7dd5c3163f2cd0cb`. Projects with binding redirects or an explicit fully-qualified assembly reference to `IgniteUI.Blazor.Lite` need updating.
```
damyanpetev:
```suggestion
**Binary compatibility:** shipped assemblies are now strong-name signed. This changes the assembly identity, so `PublicKeyToken` moves from null to `7dd5c3163f2cd0cb`. Normal NuGet consumers that rebuild should require no source changes, but precompiled dependents, binding redirects, and explicit fully-qualified assembly references may need updating.
```

CHANGELOG.md
```diff
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- The release workflow is split into isolated build, signing, packaging, SBOM, publish, and release-attachment jobs with least-privilege permissions and digest-verified handoffs between them. The Key Vault credential and the publish credential are no longer available to the same job as the build step.
```
damyanpetev:
```suggestion
```

damyanpetev (commented):
LGTM for the most part, though @turbobobbytraykov I strongly question the existance of bits like .github/scripts/Publish-NuGetPackage.ps1

I know an agent will happily spew out tons of validation scripts (future maintainers be damned), but truly that one seems off if we can't even call `dotnet publish` in the yaml directly at this point. Version override should be a feature of the publishing source and/or an option of the comment IMO. It's getting a bit silly in here :)

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Every release now publishes an SPDX 2.2 SBOM, an SPDX 3.0 SBOM, and a CycloneDX SBOM covering the NuGet and npm dependencies the package actually ships, plus three Sigstore attestations — build provenance, the SPDX SBOM, and the CycloneDX SBOM — each bound to the SHA-256 digest of the signed package. All of it is attached to the GitHub release next to the package and its checksum. Verify with `gh attestation verify <package>.nupkg -R IgniteUI/igniteui-blazor`.
+- Dependency vulnerability scanning. Pull requests are gated by a dependency review that fails on a new High or Critical advisory; every release additionally scans the NuGet and npm dependencies it actually ships and attaches the report to the GitHub release.
```
damyanpetev:
```suggestion
- Еvery release additionally scans the NuGet and npm dependencies it actually ships and attaches the report to the GitHub release.
```

damyanpetev:
> > > @turbobobbytraykov Are you going to include a different release pipeline for `IgniteUI.Blazor.Templates`, or is this coming later (not part of this PR)?
> > 
> > 
> > @damyanpetev , say the word.
> 
> This PR is already large enough. Let's leave this for a separate PR.

@kdinev def separate, though we still need to discuss if those will remain in this repo and how we'll setup the release process between out-of-sync packages.

damyanpetev (approved):
This should be good enough for now.

Issue #256, opened 2026-07-17
Audit and update the IgniteUI.Blazor bootstrap script for current hook support


damyanpetev:
### Component / Area

Other

### Is your feature request related to a problem?

The bootstrap script used by IgniteUI.Blazor may still rely on outdated or deprecated hooks and may not include all behavior needed after the recent cleanup of ignored JS hooks. This creates risk that package initialization is incomplete or diverges from the runtime expectations of the full product integration.

### Describe the solution you'd like

Review and update the IgniteUI.Blazor (full package and Lite if needed still)  bootstrap script so it reflects the current supported initialization hooks and required startup behavior.

Scope should include:
- auditing the bootstrap path for deprecated or stale hook usage
- comparing the current bootstrap flow with the runtime behavior expected after the JS hook cleanup work in PR #192
- restoring any missing functionality required for package initialization, resource loading, or runtime hookup
- aligning the bootstrap script with the modernized interop and loader architecture introduced by related follow-up tasks
- validating that the script behaves correctly in supported hosting scenarios

### Proposed API or Usage

```csharp
// Illustrative checks
- remove deprecated startup hooks
- ensure bootstrap imports/registers the current interop module(s)
- verify startup callbacks and load sequencing still occur in the expected order
```

### How important is this feature to you?

Critical — blocking my project

### Additional context

Related public context:
- PR #192 cleaned up ignored browser-warning-related JS hooks and is a useful comparison point for bootstrap expectations: https://github.com/IgniteUI/igniteui-blazor/pull/192

PR #404, opened 2026-09-09
feat: vite/ESM build, zero-hook initializer, public `api.js` module

### LINKED ISSUE #233: Ignite UI for Blazor stops working with .NET 9 Assets helper (no error but components not registered)
### Description

When using the .NET 9 Assets helper (static asset fingerprinting) to load `app.bundle.js`, Ignite UI for Blazor stops working.

No error is shown, but custom elements are not registered and components do not function.

### Hosting Model

Blazor Server

### .NET Version

9.0, 10.0

### Ignite UI for Blazor Version

26.1.51

### Component / Area

Other

### Browser

Edge

### Operating System

Windows

### Steps to Reproduce

1. Follow the [official documentation](https://www.infragistics.com/products/ignite-ui-blazor/blazor/components/general-getting-started#net-6-and-later-applications) and add the script reference in the fallback page:

```razor
<!-- Components/App.razor -->
<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>
```

2. Change it to use the .NET 9 Assets helper:

```razor
<!-- Components/App.razor -->
<script src="@Assets["_content/IgniteUI.Blazor/app.bundle.js"]"></script>
```

3. Run the app and use any Ignite UI component (for example IgbCard)
4. Open the browser console and execute:

```
customElements.get("igc-card")
```

### Actual Result

- No error is displayed
- Components do not work
- Custom elements are not registered
- The above call returns undefined

Errors may appear later when using event handling or data binding.

### Expected Result

Ignite UI for Blazor should work correctly when app.bundle.js is loaded via the .NET 9 Assets helper with fingerprinting enabled.

## Cause

The implementation in app.bundle.js assumes a fixed file name and does not account for fingerprinted file names.

## Solution

Remove the script reference from the fallback page:

```razor
<!-- Components/App.razor -->

<!-- 👇️ Remove this line -->
<script src="@Assets["_content/IgniteUI.Blazor/app.bundle.js"]"></script>
```

Ignite UI for Blazor works correctly without explicitly loading this file.

### Additional Details

Ignite UI for Blazor includes a JavaScript initializer:
```
IgniteUI.Blazor.{fingerprint}.lib.module.js
```

This initializer is automatically loaded during Blazor startup.
It performs the same role as app.bundle.js by dynamically loading app.bootstrap.js.
Because of this:

- app.bundle.js is effectively redundant
- Explicitly loading it can break behavior when fingerprinting is applied

## Suggested Changes

- Remove the instruction to load app.bundle.js from the [documentation](https://www.infragistics.com/products/ignite-ui-blazor/blazor/components/general-getting-started#net-6-and-later-applications)
- Add a note such as: ".NET 6 and later do not require explicit loading of app..bundle.js"

## Possible Improvement

Since IgniteUI.Blazor.{fingerprint}.lib.module.js is already fingerprinted:

- It could directly reference app.{fingerprint}.bundle.js
- This would eliminate the need for app.bootstrap.js
- It would reduce dynamic script injection and improve startup performance

As a further improvement, the initializer could include the logic currently in app.{fingerprint}.bundle.js, removing the need for runtime script insertion entirely.

### Reproduction URL

_No response_

### Attachments

I attach a .NET 10 Blazor Web App project (Interactive = Server, Global) that reproduce the problem I reported in this issue. See the code and comment inside the Components/App.razor.

[📦BlazorApp1.zip](https://github.com/user-attachments/files/29323391/BlazorApp1.zip)

damyanpetev:
Fixes #233. Addresses the bootstrap audit in #256.

> NOTE: ℹ️ Committed in separate chunks for review

## Why

- The `app.bundle.js` → `app.bootstrap.js` → `app.<hash>.bundle.js` chain located its chunks by regexing its own script URL. Under .NET 9+ fingerprinting (`@Assets[...]`) that silently no-ops and the app renders blank (#233).
- The script tag itself has been redundant since the JS initializer gained the Blazor Web App hooks: the bundle already auto-loads on every hosting model.
- `igRegisterScript` was a window global with no defined safe-point, and a script resolving before registration was dropped silently.

## What changed

**Build: webpack → vite, native ESM output.**
- Chunks are ES modules resolved relative to their importer, so no runtime filename inspection exists anymore — the structural fix for #233.
- Chunk shape matches the webpack output: one `igniteui-core` chunk (runtime plus all component metadata), our interop code as `app`, `igniteui-webcomponents` as the single lazy chunk. A commented knob in `vite.config.mts` splits the metadata into lazy per-component chunks instead.
- lit-html ships as a fixed-name module with its real exports (`_content/IgniteUI.Blazor/lit-html.js`), and every lit-html import in the graph goes through it, so an app mixing lit-based libraries can dedupe to one copy with a single import-map entry.
- Third-party license comments stay inline in the chunks, and vite's `build.license` emits `THIRD-PARTY-LICENSES.md` alongside the static assets.
- `tsc --noEmit` joins the build as the type-check gate (vite only transpiles), with `types: []` so Node-only globals are compile errors. That surfaced one real bug: the in-process (WASM) data path referenced Node's `global`, which webpack had polyfilled; it is `globalThis` now.
- `npm run build` produces the complete `wwwroot`, themes included; the separate `copythemes` script and CI step are gone. `webpack.config.js`, the webpack dependencies, `public_path.ts` and the undocumented `igPublicPath` / `InfragisticsBlazor` globals are removed.

**Loading: the JS initializer is the sole loader.**
- `IgniteUI.Blazor.Lite.lib.module.js` is generated as a flat list of static imports covering the app entry's closure and `api.js`. It exports no hooks: Blazor awaits the module import, the module system awaits the imports. The graph loads in one hop instead of three.
- A missing bundle fails Blazor startup loudly with the failing URL, by design.
- `app.bootstrap.js` is deleted. `app.bundle.js` remains as a static shim for existing tags: it loads nothing and queues `igRegisterScript` / `igRemoveScript` calls made while the page parses, which `api.js` replays when it loads.

**Public ES module `_content/IgniteUI.Blazor/api.js`** (typings in `api.d.ts`):
- `registerScript(name, fn, shouldCall = false)` / `removeScript(name)`. The default flips from the global's `true`: every `*Script` parameter in Lite except the value-supplying `DataScript` / `ItemsScript` wants the function itself, and every example (ours and the public grid docs) passed `false` explicitly. The deprecated global keeps `true`.
- `html`, lit-html's template tag, the same instance the components render with.
- The window globals keep working but warn once per API with the module import to use instead.
- A `*Script` name that is not registered when the component resolves it now logs a warning naming the script.

**Docs and tests.**
- Setup guidance drops the script tag for this package; the `@Assets` warning is scoped to the full product, whose loader still breaks under fingerprinting. `*Script` registration moves out of setup into its own components-skill reference (`client-scripts.md`). All `*Script` examples, including the 137 XML doc comments, use the module import, and every in-repo host follows the no-tag path.
- `npm test` (`node --test`, run in CI after the asset build) pins the built `wwwroot` invariants: no URL inspection, a plain-import initializer, real exports on `api.js` and `lit-html.js`, no internals in `api.d.ts`, legal notices present, the two `shouldCall` defaults, and the legacy stub's queue behaviour via `node:vm`.

## Compatibility

- No C# or wire-format changes; the interop contract suite passes unchanged.
- Existing apps need no changes: the tag keeps working, including wrapped in `@Assets[...]`, and the deprecated globals keep working with a console notice.
- Registering a `*Script` after its component rendered remains unsupported (now warns); order-independent registration is a separate follow-up.
- Re-hosting the assets (CDN/self-host), the one job `igPublicPath` could have done, is a standard [import map](https://learn.microsoft.com/en-us/aspnet/core/blazor/fundamentals/static-files?view=aspnetcore-10.0#importmap-component) prefix entry such as `{"imports": {"./_content/IgniteUI.Blazor/": "https://cdn.example.com/ig/"}}`, which relocates the whole graph, initializer included.

## Verification

- Auto-load confirmed on Blazor Server, standalone WASM (trimmed publish, fingerprinted initializer in the boot config), Blazor Web App (all render modes), and BlazorWebView (.NET 10 WinForms runtime probe).
- Legacy path smoke: stub queue replay, parse-time-captured `igTemplating.html` at render time, one deprecation warning per API.
- Blazor Web App host: tag / no-tag / `@Assets`-tag matrix and module-registration cases confirmed.
- Full .NET test suite and `npm test` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/IgniteUI.Blazor.Lite.TestBed/wwwroot/app.js
```diff
@@ -142,13 +142,10 @@ function Spy(obj, method) {
...
-    function TmplHandler(ctx) {
-      var html = window.igTemplating.html;
-      return html`<div>Template</div>`;
-    },
-    false,
-  );
+// Classic script, so the public module is reached with a dynamic import (already loaded by the initializer).
+async function generateClientTmpl(name) {
+  const { registerScript, html } = await import('/_content/IgniteUI.Blazor/api.js');
+  registerScript(name, function TmplHandler(ctx) {
```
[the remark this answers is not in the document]
damyanpetev: Switched every api.js example to `./_content/IgniteUI.Blazor/api.js` (docs, the 137 XML doc comments, TestBed), with a note that the specifier resolves against the importing script's URL.

One correction to the premise: import maps match on the resolved URL, so at the site root `/_content/...` and `./_content/...` hit the same prefix entry. The absolute form only breaks under a path base, where it is wrong regardless of import maps, which is reason enough to prefer `./`. `document.baseURI` is not needed: a classic script's `import()` already resolves `./` against the document base, and Blazor's JS `import` interop does the same.

src/src/app.bundle.ts
```diff
@@ -0,0 +1,51 @@
...
+  const stubTemplating = {
+    get html(): TemplateTag {
+      return (strings, ...values) => {
+        const current = (window as unknown as LegacyWindow).igTemplating;
+        if (!current || current === stubTemplating) {
+          throw new Error(
+            '[Ignite UI] the templating API is not loaded yet. Import { html } from ' +
+              "'_content/IgniteUI.Blazor/api.js', or call the template only after the components render.",
+          );
+        }
```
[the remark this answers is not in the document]
damyanpetev: Fixed: both messages now use the documented `./_content/IgniteUI.Blazor/api.js` form.

src/src/api.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const warned = new Set<string>();
+
+function deprecated(api: string, replacement: string): void {
+  if (warned.has(api)) {
+    return;
+  }
+  warned.add(api);
+  console.warn(`${api} is deprecated — import { ${replacement} } from './_content/IgniteUI.Blazor/api.js' instead.`);
+}
```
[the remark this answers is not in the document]
damyanpetev: Keeping the message as is. `./_content/IgniteUI.Blazor/api.js` is the form every example uses, and the skill reference explains when a subfolder script needs `../`. A console notice is not the place for resolution rules; the message points at the module, the docs cover placement.

src/src/api.ts
```diff
@@ -0,0 +1,110 @@
...
+const w = window as unknown as LegacyWindow;
+const stubQueue = w.igRegisterScript?.__igQueue;
+
+if (stubQueue) {
+  // Replay through the shims so queued legacy calls get the deprecation notice too.
+  for (const { name, func, shouldCall } of stubQueue) {
+    registerScriptShim(name, func, shouldCall);
+  }
+  for (const name of w.igRemoveScript?.__igQueue ?? []) {
+    removeScriptShim(name);
```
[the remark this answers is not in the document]
damyanpetev: Correct in the abstract, but the only order that goes wrong is removing a name before anything registered it and then registering it, all during page parse, through the deprecated globals. Register-then-remove replays correctly. Not adding an ordered op queue to a compatibility shim for that.

package.json
```diff
@@ -2,10 +2,9 @@
   "name": "igniteui-blazor-lite",
   "private": true,
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "cross-env NODE_ENV=production node node_modules/webpack/bin/webpack.js --mode production --config ./webpack.config.js  --color --bail",
-    "build:dev": "cross-env NODE_ENV=development node node_modules/webpack/bin/webpack.js --mode development --config ./webpack.config.js  --color --bail",
-    "copythemes": "ncp node_modules/igniteui-webcomponents/themes ./src/wwwroot/themes",
+    "test": "node --test \"tests/js/**/*.test.mjs\"",
+    "build": "tsc --noEmit -p tsconfig.json && vite build && tsc -p tsconfig.api.json",
+    "build:dev": "vite build --mode development && tsc -p tsconfig.api.json",
```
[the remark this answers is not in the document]
damyanpetev: Added the same `tsc --noEmit` gate to `build:dev`.

skills/igniteui-blazor-components/references/client-scripts.md
```diff
@@ -0,0 +1,100 @@
...
+
+registerScript('OnDialogClosing', (evt) => {
+    if (!confirm('Discard changes?')) {
+        evt.preventDefault();
+    }
+});
+```
+
+- The specifier is relative to the importing script's own URL: `./_content/…` from a script at the wwwroot root, `../_content/…` from a subfolder. A classic script's `import()` resolves it against the document base.
+- **Register at the top level of a `type="module"` script, or any time before Blazor starts.** Module scripts are deferred, so the tag's position in the page does not matter. Registering after a component has rendered is not supported ATM — the library logs a console warning naming the unresolved script.
```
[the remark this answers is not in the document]
damyanpetev: Tightened the wording: the bullet now explains why a page-level module script is in time (Blazor awaits the component bundle before its first render, so its startup chain is strictly longer than a deferred script) and names the app's own JS initializer as the hard guarantee for anyone who wants one. Not documenting a manual startup gate as the recipe: a miss is reported with a console warning, and order-independent late registration is the planned fix, which makes the gate unnecessary.

skills/AGENTS.md
```diff
@@ -58,7 +58,7 @@ You are an expert in C#, Blazor, and scalable web application development. You w
...
-- **Setup**: `builder.Services.AddIgniteUIBlazor()` in `Program.cs`, `@using IgniteUI.Blazor.Controls` in `_Imports.razor`, one theme stylesheet, and `_content/IgniteUI.Blazor/app.bundle.js` before the Blazor framework script in the host page. Missing the script tag renders the app blank.
+- **Setup**: `builder.Services.AddIgniteUIBlazor()` in `Program.cs`, `@using IgniteUI.Blazor.Controls` in `_Imports.razor`, and one theme stylesheet in the host page. `IgniteUI.Blazor.Lite` loads its scripts itself (JS initializer); the full `IgniteUI.Blazor`/`.Trial` also needs `_content/IgniteUI.Blazor/app.bundle.js` before the Blazor script in Blazor Web Apps — never wrapped in `@Assets[...]`.
```
kdinev, line 61: Explain why the licensed package needs it. Just a short reason.

skills/igniteui-blazor-grids/SKILL.md
```diff
@@ -61,7 +61,7 @@ Grids are **not** included in `IgniteUI.Blazor.Lite`.
 | Using directive | `@using IgniteUI.Blazor.Controls` in `_Imports.razor` |
 | CSS — full grids | **both** `_content/IgniteUI.Blazor/themes/light/bootstrap.css` **and** `_content/IgniteUI.Blazor/themes/grid/light/bootstrap.css` |
 | CSS — Grid Lite | `_content/IgniteUI.Blazor.GridLite/css/themes/light/bootstrap.css` only — not the two above |
-| Script | `_content/IgniteUI.Blazor/app.bundle.js` before the Blazor framework script |
+| Script | None for `IgniteUI.Blazor.GridLite`; full grids (`IgniteUI.Blazor`/`.Trial`) need `_content/IgniteUI.Blazor/app.bundle.js` before the Blazor framework script in Blazor Web Apps — never wrapped in `@Assets[...]` |
```
kdinev, line 64: This place also needs a short answer to "Why?".

skills/igniteui-blazor-generate-from-image-design/SKILL.md
```diff
@@ -45,7 +45,7 @@ Start every region with the most appropriate Ignite UI component. Fall back to p
 
 There are no separate DV packages. If a required package is missing from the project, identify the right package and version and **ask before editing the `.csproj`**.
 
-Register every `Igb*Module` you use in `Program.cs`, add `@using IgniteUI.Blazor.Controls` to `_Imports.razor`, and confirm the theme stylesheet and `app.bundle.js` are in the host page — see the components skill's [`setup.md`](../igniteui-blazor-components/references/setup.md).
+Register every `Igb*Module` you use in `Program.cs`, add `@using IgniteUI.Blazor.Controls` to `_Imports.razor`, and confirm the theme stylesheet is in the host page (on `IgniteUI.Blazor`/`.Trial` in a Blazor Web App, also the `_content/IgniteUI.Blazor/app.bundle.js` script tag before the Blazor script; `IgniteUI.Blazor.Lite` needs no tag) — see the components skill's [`setup.md`](../igniteui-blazor-components/references/setup.md).
```
kdinev, line 48: This differs from the explanation in our dotnet/skills skill: https://github.com/dotnet/skills/pull/1111/changes#diff-6fbe289f6bfc3d6caa7652f7b348ca92b9af9e27c0dd5b3f6ce96cc49794a22bR44

In this one: 
1. List the modules in the licensed package that components don't function without.
2. For the rest, explain that the modules are for eager loading, but if not includes would be lazy-loaded.

skills/igniteui-blazor-components/SKILL.md
```diff
@@ -41,13 +42,14 @@ All four use the `IgniteUI.Blazor.Controls` namespace and serve static assets fr
 ## Rules that apply to every component
 
 - **Registration.** `builder.Services.AddIgniteUIBlazor()` in `Program.cs` is required. Passing `typeof(Igb<Name>Module)` arguments eagerly pre-loads exactly those modules; with no arguments every module is available. In `IgniteUI.Blazor.Lite` each component also registers its own module on first render, so the explicit list is a bundle-size optimization rather than a correctness requirement.
-- **Runtime script.** `<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>` must appear before the Blazor framework script in the host page. Missing it means no web components register and the app renders blank.
+- **Runtime script.** `IgniteUI.Blazor.Lite` loads its component bundle itself (JS initializer, every hosting model). The full `IgniteUI.Blazor`/`IgniteUI.Blazor.Trial` needs `<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>` before the Blazor framework script in **Blazor Web Apps** — never wrapped in `@Assets[...]` (fingerprinting it renders the app blank).
```
kdinev, line 45: Is this applicable when not using a grid or chart component, since we haven't yet ripped the rest of the components out of `IgniteUI.Blazor` and haven't made `IgniteUI.Blazor.Lite` a dependency yet?

skills/igniteui-blazor-components/references/setup.md
```diff
@@ -62,14 +62,24 @@ Add it to both `_Imports.razor` files in split Blazor Web App solutions.
...
 <link href="_content/IgniteUI.Blazor/themes/light/bootstrap.css" rel="stylesheet" />
 ...
-<script src="_content/IgniteUI.Blazor/app.bundle.js"></script>
 <script src="_framework/blazor.web.js"></script>   <!-- or blazor.server.js / blazor.webassembly.js / blazor.webview.js -->
 ```
 
-Both tags are required: without the stylesheet components render unstyled, without `app.bundle.js` they do not render at all. `app.bundle.js` must come **before** the Blazor framework script.
+**`IgniteUI.Blazor.Lite` (≥ 0.1.0) loads its component bundle itself** — a JS initializer runs during Blazor startup on every hosting model: Blazor Server, standalone WASM, Blazor Web App (`blazor.web.js`, any render mode), and BlazorWebView/Hybrid.
+
+**The full product (`IgniteUI.Blazor` / `IgniteUI.Blazor.Trial`, ≤ 26.1.x) still needs the tag in Blazor Web Apps** — it ships a classic-only initializer that `blazor.web.js` ignores. There the tag must come **before** the Blazor framework script; classic Blazor Server and standalone WASM apps on the full product may omit it.
```
kdinev, line 75: So, if I get this correctly, we will drop this requirement in `26.2`?

skills/igniteui-blazor-components/references/setup.md
```diff
@@ -62,14 +62,24 @@ Add it to both `_Imports.razor` files in split Blazor Web App solutions.
 
 Host page is `wwwroot/index.html` (WASM/MAUI), `Pages/_Host.cshtml` (Server), or `Components/App.razor` (Web App).
 
+The theme stylesheet is the only required tag — without it components render unstyled.
```
kdinev, line 65: This statement _The theme stylesheet is the only required tag_ goes in conflict with the next statement that `IgniteUI.Blazor` also requires a `script` tag with the `app.bundle.js` referenced.

skills/igniteui-blazor-components/references/client-scripts.md
```diff
@@ -0,0 +1,100 @@
...
+    return html`
+        <div style="display: flex; align-items: center; gap: 8px;">
+            <igc-avatar shape="circle" src="/agent.jpg"></igc-avatar>
+            <b>Support Agent</b>
+        </div>
+    `;
+});
+```
+
+`html` is the Lit HTML template tag: `${}` interpolates values, `@click=${fn}` binds listeners, and so on. Ignite UI web components (`<igc-avatar>`, `<igc-icon-button>`, …) work inside a template; if the component is not used as a Blazor component elsewhere in the app, register its module explicitly with `AddIgniteUIBlazor(typeof(IgbAvatarModule), …)` in `Program.cs`.
```
kdinev, line 84: Why should I register its module explicitly?

kdinev (changes requested):

kdinev (dismissed):
This commit all looks good.

kdinev (changes requested):
I've submitted multiple review comments. Please take a look.

jsakamotoIGJP:
@damyanpetev Thank you for this work!
Registering scripts as an ES module, without touching the global namespace, is a long-awaited improvement.

One problem. A user script module loads asynchronously, so registerScript can run after a component has already rendered, and the current implementation never uses the registered script.

I will attach a zip of a small Blazor WebAssembly project (.NET 10). 
- [📦IgbLiteWasm10.zip](https://github.com/user-attachments/files/32186746/IgbLiteWasm10.zip)

Its "wwwroot/scripts.js" waits 2 seconds before calling registerScript to simulate a script-loading delay.

```js
import { registerScript, html } from "./_content/IgniteUI.Blazor/api.js";

// Emulate something going wrong while loading the script...
await new Promise(resolve => setTimeout(resolve, 2000));

registerScript("ComboItemTemplate", (args) => {
    return html`<div>${args.item.Name} (${args.item.Description})</div>`;
});
```

On this sample app, the `IgbCombo` with `ItemTemplateScript="ComboItemTemplate"` renders without the template (depends on machine specs.) The button recreates the combo, and then the template works, so the order is the only issue.

In practice, the components start later than the user script, so this will almost never happen. Still, I would like to ask you to consider the order-independent registration follow-up. If order doesn't matter, these scripts can serve another useful purpose. I will share that another time.

damyanpetev:
> In practice, the components start later than the user script, so this will almost never happen. Still, I would like to ask you to consider the order-independent registration follow-up. If order doesn't matter, these scripts can serve another useful purpose. I will share that another time.

@jsakamotoIGJP Indeed, I believe there's an explicit note in the PR descr regarding this:

> * Registering a `*Script` after its component rendered remains unsupported (now warns); order-independent registration is a separate follow-up.

That's is what you mean, right? It's absolutely known to me and flagged for an update, but that will require some rework of the interop internals, that I'd rather not touch in this PR as well. At least now it should log a message so there's _some_ feedback what it went wrong as a temp compromise.

jsakamotoIGJP:
@damyanpetev Oh, I'm sorry, I totally overlooked the part of the PR description. I completely agree with your direction, as "order-independent registration is a separate follow-up." Again, sorry for disrupting you.

skills/igniteui-blazor-generate-from-image-design/SKILL.md
damyanpetev: Like we discussed, this entire section is superbly sus to me, since it re-iterates the setup.md from the components skill (which is weird) and now we don't have that explained in a single place in one way and the references used throughout skills are questionable at best (agent tried to make this one gracefully fallback by noting it's from a separate skill at least).

skills/igniteui-blazor-components/SKILL.md
damyanpetev: Yup, once I manage to apply the same mechanics to the other package (with the fix made months ago to the hooks for the library module initializer) to the full package, it'll be equalized (may/may not be directly related to the dependency part)

skills/igniteui-blazor-components/references/setup.md
damyanpetev: hopefully, but won't put it writing just yet

kdinev (dismissed):

CHANGELOG.md
```diff
@@ -9,18 +9,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Every release now publishes an SPDX 2.2 SBOM, an SPDX 3.0 SBOM, and a CycloneDX SBOM covering the NuGet and npm dependencies the package actually ships, plus three Sigstore attestations — build provenance, the SPDX SBOM, and the CycloneDX SBOM — each bound to the SHA-256 digest of the signed package. All of it is attached to the GitHub release next to the package and its checksum. Verify with `gh attestation verify <package>.nupkg -R IgniteUI/igniteui-blazor`.
 - Еvery release additionally scans the NuGet and npm dependencies it actually ships and attaches the report to the GitHub release.
 
 ### Changed
 
+- The client build now emits native ES modules; the package's JS initializer loads the whole script graph during Blazor startup, so no `<script src="_content/IgniteUI.Blazor/app.bundle.js">` tag is needed on any hosting model (Blazor Server, standalone WASM, Blazor Web App, BlazorWebView). Existing tags keep working — including wrapped in `@Assets[...]`.
```
skrustev, line 18: Maybe add that if the user removes it, the new `api.js` module needs to be used instead for templates.

src/src/api.ts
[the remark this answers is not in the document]
skrustev: @damyanpetev Maybe this should be noted somewhere as a bad practice since pretty much the following case will not work as expected, even though it's weird and I can't think of reasonable cases where one would use the `removeScript` on load anyway but:
```js
  window.igRegisterScript('Gone', queued, false);

  //...
  window.igRemoveScript('Gone');

  window.igRegisterScript('Gone', queued, false);
```

This will results in the script not present at all.

skrustev (dismissed):

CHANGELOG.md
damyanpetev: IIRC, the window shims (marked as deprecated) will still be there, but these will load with the library, so there could be a timing bit to contend with and it might be worth noting the new approach is preferred again.
Its also technically the "Added" section bullet, but I see the point as this being not quite clear enough.

Would be nice if the added section about the api.js had a snippet example of usage (plenty of those in the skills already) and that that is now the preferred way.
And yes then this section can point to that for registering `*Script` properties without the app.bundle.js that ensures their global versions (deprecated) it's suggested to use the api.js module. Thought that's technically repeating the point of the two other notes, but sure, if it makes it clearer.

@copilot+gpt-5.6-sol can you suggest edits to the changelog notes to clarify them based on that

skrustev (approved):

PR #406, opened 2026-09-10
Replace unicode_hack.js w/ regex uni chars, ship THIRD-PARTY-LICENSES.md at the package root

damyanpetev:
## Why

- `igniteui-core/stringExtended.ts` carried `unicode_hack.js`, a 2010-era polyfill that emulated `\p{…}` regex escapes with hand-generated Unicode 6.x character tables. Native property escapes have been in every supported engine since ES2018, and the tables were stale (hundreds of misclassified BMP code points per category, no astral coverage).
- The license manifest that #404 introduced was written into `wwwroot`, so it was packed and served as a static web asset. It is package documentation, not something a page should fetch.

## What changed

- The four character-class helpers (`isLetter`, `isLetterOrDigit`, `isDigit`, `isNumber`) are one-line `/\p{…}/u` tests; the tables, the translator, `unicode_hack` and `netRegexToJS` are deleted. The core chunk shrinks by about 11 KB gzip and the in-source legal notice is gone, so the `npm test` assertion that guarded it has nothing left to check.
- `THIRD-PARTY-LICENSES.md` moves to the NuGet package root next to LICENSE and README. Rolldown only emits inside `outDir`, so the vite build moves the file next to the csproj in `closeBundle`, and the csproj packs it when present. `.gitignore` covers it like `wwwroot`.

## Verification

- `npm run build`, `npm test`, and the .NET suite pass.
- `dotnet pack` listing: the manifest is at the package root and absent from `staticwebassets/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MayaKirova (approved):
LGTM

PR #421, opened 2026-09-14
docs: Add XML docs to the public enums

### LINKED ISSUE #417: Document component-specific public API members

damyanpetev:
## What

Adds XML documentation to all 50 public enums in `src/components/Blazor`: a summary on each type and on each member. None of them had any, so every one produced CS1591 warnings for the library's documentation file and showed up blank in IntelliSense and the API reference.

## How the text was derived

Each enum was traced to the parameters that use it and to the matching string union and rendering code in `igniteui-webcomponents`, so the member descriptions state what the value does rather than restating its name. For example, `TreeSelection.Cascade` describes the indeterminate parent state, `ButtonGroupSelection.SingleRequired` describes that clicking the selected button leaves it selected, `TextareaResize` maps to the CSS resize behavior, and `StepperTitlePosition.Auto` says what it resolves to per orientation.

Shared enums (`ContentOrientation`, `StyleVariant`, `PopoverPlacement`, `WeekDays`, `ToggleLabelPosition`, …) are described generically and name their consumers. Enums whose values are HTML attribute values (`ButtonBaseType`, `ButtonBaseTarget`, `InputType`, `TextareaWrap`) say so.

Where a value is laid out with logical CSS properties in the web component (carousel indicators, splitter panes, nav drawer edges), the docs use start/end and add the left-to-right meaning in parentheses. `PopoverPlacement.Left`/`Right` stay physical because the underlying placements are.

## Verification

`dotnet build src/IgniteUI.Blazor.Lite.csproj` succeeds on all three target frameworks with no warning pointing at any of the 50 files and no unresolved `cref` anywhere. The same descriptions were ported as JSDoc to the corresponding type aliases in `igniteui-webcomponents` in a separate change.


src/components/Blazor/MaskInputValueMode.cs
```diff
@@ -1,8 +1,13 @@
 namespace IgniteUI.Blazor.Controls
 {
+    /// <summary>
+    /// Format in which an <see cref="IgbMaskInput"/> exposes its value.
+    /// </summary>
     public enum MaskInputValueMode
     {
+        /// <summary>Only the characters typed by the user, for example <c>5551234567</c>.</summary>
         Raw,
+        /// <summary>The value with the mask literals applied, for example <c>(555) 123-4567</c>.</summary>
```
damyanpetev:
```suggestion
        /// <summary>Returns clean input , e.g. <c>5551234567</c>.</summary>
        Raw,
        /// <summary>Returns with mask formatting, e.g. <c>(555) 123-4567</c>.</summary>
```

src/components/Blazor/MaskInputValueMode.cs
```diff
@@ -1,8 +1,13 @@
 namespace IgniteUI.Blazor.Controls
 {
+    /// <summary>
+    /// Format in which an <see cref="IgbMaskInput"/> exposes its value.
+    /// </summary>
     public enum MaskInputValueMode
     {
+        /// <summary>Returns clean input , e.g. <c>5551234567</c>.</summary>
```
damyanpetev:
```suggestion
        /// <summary>Returns clean input, e.g. <c>5551234567</c>.</summary>
```

src/components/Blazor/BadgeShape.cs
```diff
@@ -1,8 +1,13 @@
 namespace IgniteUI.Blazor.Controls
 {
+    /// <summary>
+    /// Shape of an <see cref="IgbBadge"/>.
+    /// </summary>
     public enum BadgeShape
     {
+        /// <summary>Fully rounded ends.</summary>
         Rounded,
+        /// <summary>Square corners.</summary>
```
kdinev:
```suggestion
        /// <summary>Square with sharp corners.</summary>
```

Making this consistent with the `AvatarShape.Square`

kdinev (dismissed):
Just one small suggestion added for consistency.

src/components/Blazor/BadgeShape.cs
damyanpetev: Ah, unfortunately not the same thing since there are no two square options for this one and by the looks of them not sharp in this option either https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/badge#shape. Will adjust

dkamburov (approved):

PR #433, opened 2026-09-16
test: Refactor and remove unnecessary tests that produce BL0005 warnings.

### LINKED ISSUE #431: Refactor tests that set param to direct instances and remove those that are not needed
Some tests assign params over to direct instances (no-framework unit tests) - those either can be re-written with bUnit and/or some of the overlapping coverage with interop contracts removed;

Most likely a lot of these are already covered in the contract tests.

MayaKirova:
Closes #431


kdinev (dismissed):

kdinev (approved):
All looks good now. I applied a change to the testing skills.

PR #434, opened 2026-09-16
Add JsonSerializable base class. Update related types to use it instead that produce BL0005 warn.

### LINKED ISSUE #432: Refactor types like option types to use the plain object serialization
Types (usually for options/method API) that, due to shared serialization infra, inherit from component base and as such trigger those warnings.

Those should be moved to the simpler serialization path (like chat options IIRC) or made clean type on interop rework

skrustev:
Closes #432


src/components/Blazor/ChatOptions.cs
```diff
@@ -221,24 +211,14 @@ public bool AdoptRootStyles
...
-                if (value != null)
-                {
-                    this.AttachChild(value);
-                }
                 this._renderers = value;
             }
-
         }
 
         internal override void SerializeCore(RendererSerializer ser)
```
damyanpetev, line 224: I believe the point to this was to more/less convert this to JsonSerializable.Serialize instead and potentially avoid the entire `BaseJsonSerializable` (which now adds new warnings btw) and extra plumbing for `RendererSerializer`.

The base control handling for `BlazorPlainObject` didn't work for these?

damyanpetev: Also a little scary some of those were used as event payload and the current machinery relied on specific "component" (🥸) methods for that...

PR #439, opened 2026-09-17
Hide the runtime-only API of the base classes and the runtime service

### LINKED ISSUE #419: Cleanup internal infrastructure and automation leftover API left as public
- [x] Cleanup all `SetNativeElement` methdos #410
- [x] Cleanup all custom-element style [dis-]connected callback methods #411
- [x] Cleanup dead/orphaned code #310
- [x] Internalize extra module methods beyond `Register` #425

damyanpetev:
The base classes and the runtime service exposed the interop plumbing as public or protected members: serialization helpers, ref-change callbacks, the JavaScript callback object, module-cache bookkeeping, the data-source manager and its adapters. None of it is documented or used by the samples and docs of either product, and all of it serves the client runtime the interop rewrite replaces. Lite also shares the `IgniteUI.Blazor.Controls` namespace with the full product, so every public type here is a collision waiting to happen once the full package depends on Lite.

What apps use stays public: `AddIgniteUIBlazor` and its settings, each module's `Register`, the component parameters and `EnsureReady`.

## `IIgniteUIBlazor` is a marker

The injected service has no members. Its runtime surface lives on an internal `IIgniteUIBlazorRuntime` the same class implements, reached through one `AsRuntime()` helper that throws if a foreign implementation is registered. `IgniteUIBlazor`, `WebCallback` and `ModuleLoader` are internal; the JS-invokable methods stay public on the internal callback because interop dispatch needs that. `ModuleLoader` had one public method and the client loader only knows the product's own module names, so nothing outside could load anything with it.

## Data change notifications move to the combo

The five `Notify*Item` methods and `SuspendNotifications`/`ResumeNotifications` were on every component but only mean something for one with a bound collection. They are now the documented `IDataSourceNotifications` interface, implemented by `IgbCombo<T>`, with their original parameter names so existing named-argument calls still compile. Binding an `ObservableCollection` works as before.

## Types now internal or removed

Internal: `BaseCollection<T>`, `DataIntentAttribute`, `IDataIntentAttribute`, the dynamic-content types, `SerializationContext`, `SerializationFilter`, `JsonSerializable`, `MarshalByValueFactory`. Removed: `IgbComponentRendererContainer` and its event args, which nothing referenced.

## Base-class plumbing narrowed, not opened

The bases stay public because components derive from them, but their plumbing does not have to be. A member only the component itself and its overrides touch becomes `private protected`; a member the runtime or a collection calls on another object becomes `internal`. Neither is `protected internal`, which is the union of the two and would keep a member visible to anyone deriving from the base in their own assembly.

So `private protected` now covers the cascading-parent name, the element's name field and resource-string calls, the serialization and enum-transform hooks, the property-dirty check, and the item components' cascading references to their parent. `internal` covers the dirty-marking call, which a collection makes on its parent, and child lookup by name on the element, where a collection calls it across objects; on the control that lookup is `private protected` and was public before. Helpers with no caller outside their own class are simply `private`.

Kept protected, as extension points for as long as their feature exists: `EnsureModulesLoaded` with the injected service, the method-invocation calls and the event-behavior pair. The render hooks are untouched here and follow in a separate change, together with the render split they describe.

## Breaking change note

Internal API that was unintentionally public is now internal. Code that read members off `IIgniteUIBlazor` injects `IIgniteUIBlazorSettings` for the settings and needs nothing else.

## Verification

Library, stories and TestBed build clean, and the unit suite passes on net8.0, net9.0 and net10.0. The trimmed publish smoke passes, which proves the DI container constructs the internal runtime and JavaScript still calls into the internal callback.


dkamburov (approved):
LGTM

PR #440, opened 2026-09-17
refactor: hide the render hooks of the base classes

### LINKED ISSUE #419: Cleanup internal infrastructure and automation leftover API left as public

damyanpetev:
Addition to #439 moved as separate since its a bulk repetitive change.

`UseDirectRender`, `SupportsVisualChildren`, `ResolveDisplay` and `DirectRenderElementName` decide how a component's markup is produced. Every component overrides them, only the component itself reads them, and the in/direct render distinction they describe is on its way out, so they are `private protected` now: still overridable inside the library, invisible to derived components elsewhere and to app code. The 87 files touched change one modifier per override and nothing else.

## Verification

Library, stories and TestBed build clean; the unit suite passes on net8.0, net9.0 and net10.0. No changelog entry: the Unreleased "Infrastructure" bullet from the previous PR already covers the rendering plumbing.


dkamburov (approved):
LGTM!

Issue #346, opened 2026-08-19
[JS interop] Use narrow module-scoped APIs.


MayaKirova:
Interop is routed through a shared process-wide bridge and shared reference maps rather than a narrow module-scoped API owned by the control ATM.

PR #445, opened 2026-09-18
refactor: resolve BL0005 & misc warnings, suppress BL0007

### LINKED ISSUE #427: Address BL0005 warnings in test, samples and a few lib
More details TBD
- Some tests assign params over to direct instances (no-framework unit tests) - those either can be re-written with `bUnit` and/or some of the overlapping coverage with interop contracts removed;
- The second bit is around types (usually for options/method API) that, due to shared serialization infra, inherit from component base and as such trigger those warnings:
  Image
  Those should be moved to the simpler serialization path (like chat options IIRC) or made clean type on interop rework

skrustev:
## Description
* Add internal methods that update needed components with custom internal logic. (BL0005) - Closes #427
* Clean other warnings the came up during last merges.
* Suppress BL0007 since the refactoring too big for the current scope. Related #447

 - [ ] Bug fix
 - [ ] New functionality
 - [x] Refactoring (no functional changes)
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog

## How Has This Been Tested?

 - [x] Unit tests
 - [ ] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **.NET version**:
 - **Hosting model**: 
 - **Browser(s)**:
 - **OS**:

## Checklist:

Closes #


src/componentsBase/DynamicContentHolder.cs
```diff
@@ -299,13 +299,8 @@ private void OnContextChanged(T? oldValue, T? newValue)
...
             {
                 var template = (IgbTemplateContent<T>)Component;
-
-                if (_hasPopulatedContext && Context != null)
-                {
-                    template.Context = Context;
-                }
-                template.Template = Template;
-                template.Update();
+                var context = _hasPopulatedContext && Context != null ? Context : template.Context;
```
damyanpetev: That's not equivalent of the old logic tho; That didn't set the Context at all outside the conditions.
Point being, `template.Context` has a setter and that also flips an internal `_hasPopulatedContext` which affects rendering apparently. Before the change - no flag, after - flag is now true;

Make the context assign optional in the new method instead (careful w/ optional param context, in case it can be null it might be safer to use a separate flag;

src/IgniteUI.Blazor.Lite.csproj
```diff
@@ -26,6 +26,7 @@
 
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <NoWarn>$(NoWarn);BL0007</NoWarn>
```
damyanpetev: Ah, okay, since we're bundling this into this PR - I wanted to log an issue
(Could end up being related to one of the existing interop issues #346, but not for certain) and have a comment above the disable that's it temporary until that issue is resolved.

damyanpetev (commented):
And I see a few more BL0005 in the CI before this can close #427, which I'd like to do at this point.

.editorconfig
```diff
@@ -294,6 +294,10 @@ dotnet_diagnostic.IDE2000.severity = warning
 # project generates an XML docs file; always applied in the IDE.
 dotnet_diagnostic.IDE0005.severity = warning
 
+# BL0007: Component parameters should be auto properties.
+# Note: Currently setting diagnostics in .editorconfig do not work for .razor files, even if added to filters. (https://github.com/dotnet/roslyn/issues/85310)
```
damyanpetev:
```suggestion
# REPO: Temporary down-level BL0007:Component parameters should be auto properties
# awaiting refactor <issue> 
```

And move the note to if you want to leave it on top of the file/or the 
Enforced severities section; and maybe add if needed those can go in `.globalconfig` per https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files

damyanpetev (approved):

